PR #12, opened 2018-02-04
Update 7z.txt to version 18.01

blackcrack:
let hold working programs actual :
http://www.7-zip.org/

and btw, it works..

### DIFF
--- a/7zip.txt
+++ b/7zip.txt
@@ -1,13 +1,13 @@
 [Section]
 Name = 7-Zip
-Version = 16.04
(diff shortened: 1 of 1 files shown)


blackcrack:
for this, Program announce/updating !
that is nonsense and has nothing to do with programming !
Programming in Reactos, yes, indeed, right, to know from where and who has programming, but there ?!!
nope ! You trust me or you trust me not ! I be now more as 10 Years on the side of Reactos !

PR #14, opened 2018-02-05
[WORDVIEW] Update downloads links to web.archive.org

learn-more:


learn-more:
Yeah, this is the date they were saved.
If i have the same date on all, it will just resolve to this url first before downloading, so this should be slightly faster.

PR #17, opened 2018-02-08
Add new apps: MuseScore, MySql, Photivo and USBDeView.

AS383:
Add new apps: MuseScore, MySql, Photivo and USBDeView (with Russian translation).

### DIFF
--- a/musescore.txt
+++ b/musescore.txt
@@ -1,0 +1,15 @@
+[Section]
+Name = MuseScore
+Version = 2.1
+License = GNU GPLv2
+Description = Professional open source music notation software.
+Size = 54.6 MiB
+Category = 15
+URLSite = https://musescore.org/
+URLDownload = https://ftp.osuosl.org/pub/musescore/releases/MuseScore-2.1/MuseScore-2.1.msi
+SHA1 = 4af01c0267ab0b22be2e8a5ec957903fbec68c62
+CDPath = none
+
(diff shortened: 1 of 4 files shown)


learn-more:
You tested all of them?

AS383:
Yes, I test all some time ago. If wanted I can retest with current version and populate results to wiki. 
...@mail.ru is not intentional. I think it would be resolved to normal login, but it does not.

learn-more:
@AS383 I truncated the first part of the email, because if it was not your intention to use this email address, it would still show up in the history of the comments here.
You can inspect the full email address here: https://github.com/reactos/rapps-db/commit/562105ab3380f64b9a7a812952e2443c06e813ad.patch

AS383:
@learn-more Thanks for hint. Can I avoid including e-mail in future?
Also I found sha1 sum of USB Deview is wrong. Please exclude it from the pull.

sanchaez:
@AS383 we currently require contributions with both your name and _real_ email. Tell us if you wish to use another address.

learn-more:
@AS383 We need some email, if you do not want to use your public one you can have github create a masked one for you.
To update your existing commits:
https://stackoverflow.com/a/1320317

AS383:
I does not want to specify another address. It ok to use current as public for any contributions lists. Thanks everybody for support :)

PR #18, opened 2018-03-06
Terminal "Terminate 5"

blackcrack:
It is testet and it work's in Reactos, tested with 
reactos-bootcd-0.4.9-dev-138-g1f7cb06-x86-msvc-win-dbg.7z
htp and work together ;)
http://www.terminate.com/
https://en.wikipedia.org/wiki/Terminate_(software)

### DIFF
--- a/terminate.txt
+++ b/terminate.txt
@@ -1,0 +1,116 @@
+[Section]
+Name = Terminate 5
+Version = 5.00
+License = unknown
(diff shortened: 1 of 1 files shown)


blackcrack:
for this, Program announce/updating !
that is nonsense and has nothing to do with programming !
Programming in Reactos, yes, indeed, right, to know from where and who has programming, but there ?!!
nope ! You trust me or you trust me not ! I be now more as 10 Years on the side of Reactos !

PR #27, opened 2018-05-28
Add QEMU

archeYR:
Add QEMU to rapps-db. Needs the -display sdl flag to display the screen correctly.
virtualbox_reactos_2_04_05_2018_19_58_55

### DIFF
--- a/qemu.txt
+++ b/qemu.txt
@@ -2,7 +2,7 @@
 Name = QEMU
 Version = 2.7.0
 License = GPL v2
-Description = An open source and very fast machine emulator/virtualizer. It can emulate processors based on many various architectures (eg. x86, ARM, MIPS, PowerPC). Run with -display sdl flag to avoid the display glitches.
+Description = An open source and very fast machine emulator/virtualizer. It can emulate processors based on many various architectures (eg. x86, ARM, MIPS, PowerPC). Run with "-display sdl" flag to avoid the display glitches.
 Size = 20.9 MiB
 Category = 15
 URLSite = https://www.qemu.org/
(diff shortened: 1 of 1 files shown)


qemu.txt
```diff
@@ -0,0 +1,16 @@
+[Section]
+Name = QEMU
+Version = 2.7.0
+License = GPL v2
+Description = An open source and very fast machine emulator/virtualizer. It can emulate processors based on many various architectures (eg. x86, ARM, MIPS, PowerPC). Run with -display sdl flag to avoid the display glitches.
```
binarymaster: Suggestion: You could use double quotes for `"-display sdl"`.

archeYR: Done.

learn-more:
What happens without the `-display sdl` flag?

archeYR:
@learn-more Here is what happens without that flag.
virtualbox_reactos 2_12_06_2018_22_37_27
It is almost unusable then.

learn-more:
I am not decided about this issue, due to it needing special commandline arguments for it to be usable.
Anyone care to weigh in?

karlexceed:
I feel that though you can add a note about the command flag to the program description, it should be left out of RApps until this issue is resolved.

Unless:
- You could add something to the DB file so that RApps can set that flag or generate a .BAT launcher for it.
- RApps had the ability to set that flag / generate a BAT launch file.

archeYR:
@karlexceed Including a batch file wouldn't really make a sense, as QEMU is meant to use with commandline. The notice about flag is already added to the description.

archeYR:
Also, I want to point out that there is an older version of QEMU (from 2013) which doesn't have that problem, we can include the older version for now, and then upgrade when the issue with display is resolved.

learn-more:
@archeYR That sounds like a winner.
Is there already an issue for this problem?

archeYR:
@learn-more Alright then, I will change the QEMU version soon. There is no issue made specifically for QEMU, but it seems to be similar to [CORE-14589](https://jira.reactos.org/browse/CORE-14589)

learn-more:
No worries, everyone does this in his spare time :)

PR #54, opened 2019-04-04
Added Spotify to RAPPS

BenNottelling:
Specifically Version `1.0.29.92` which is an old version that still supports XP. In my experience it's hard to find a download for versions of Spotify that work on XP, so I figured I'd host it on my server and add it to rapps for everyone to enjoy.

If only our audio stack made this practical and didn't make me cry.

### DIFF
--- a/spotify.txt
+++ b/spotify.txt
@@ -1,0 +1,13 @@
+﻿[Section]
+Name = Spotify
+Version = 1.0.29.92
+License = Freeware
+Description = Spotify is one of the most popular services for streaming music and includes both a free and paid plan. The Spotify client has dropped support for XP, but this archived version still supports it.
+Size = 47.7 MB MiB
+Category = 1
+URLSite = https://www.spotify.com/us/
+URLDownload = https://bennottelling.com/ros/apps/SpotifyFullSetup1.0.29.92.exe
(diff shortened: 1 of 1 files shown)


BenNottelling:
Here are some pictures to show it works, icons don't load right but overall it is usable (which is better than other apps in the database such as steam). Turning off the friends panel help with performance.
loggedin
login

spotify.txt
```diff
@@ -0,0 +1,13 @@
+﻿[Section]
+Name = Spotify
+Version = 1.0.29.92
+License = Freeware
+Description = Spotify is one of the most popular services for streaming music and includes both a free and paid plan. The Spotify client has dropped support for XP, but this archived version still supports it.
+Size = 47.7 MB MiB
+Category = 1
+URLSite = https://www.spotify.com/us/
+URLDownload = https://bennottelling.com/ros/apps/SpotifyFullSetup1.0.29.92.exe
```
Mouvedia: https://filehippo.com/fr/download_spotify/68194/
less likely to go down

BenNottelling: That's true, but it's not a direct link to the file, plus I've got cloudflare on my server so that should help with uptime.

PR #57, opened 2019-04-14
[RAPPS] Updating Spotify Link

BenNottelling:
For some reason the apps manager can't download the file from my server, so I put it on Archive.org, which it probably belongs on anyways. 
I uploaded it to the software repository originally so the link was going to be this https://archive.org/download/SpotifyFullSetup1.0.29.92/SpotifyFullSetup1.0.29.92.exe
but when I search the URLs in the database they used web snapshots instead of software uploads, so I followed suit, landing on
https://web.archive.org/web/20190414171907/http://bennottelling.com/ros/apps/SpotifyFullSetup1.0.29.92.exe


HBelusca:
> but when I search the URLs in the database they used web snapshots instead of software uploads [...]

That's because originally we downloaded these apps from the original website, and then later they went out, so we had to rely on the web snapshots.
In your particular case however you uploaded manually the software to archive.org, so we can directly use the download link.

PR #61, opened 2019-06-03
[RAPPS-DB] Add Microsoft Visual C++ 2015 Redistributable Update 3

oleg-dubinskiy:
vc++2015-2
vc++2015
Needed by programs which are compiled with Microsoft Visual Studio 2015. Installs fine in ReactOS.
Although installer has a bit incorrect position of the "Install" button at the start of setup, same as in Wine, it does not affect the installation efficiency.

### DIFF
--- a/vc2015run.txt
+++ b/vc2015run.txt
@@ -1,0 +1,65 @@
+[Section]
+Name = Microsoft Visual C++ 2015 Redistributable Update 3
+Version = 14.0.24212
+License = Unknown
+Description = Needed by applications compiled with Visual Studio 2015. Contains: atl140.dll, concrt140.dll, mfc140.dll, mfc140u.dll, mfcm140.dll, mfcm140u.dll, msvcp140.dll, vcamp140.dll, vccorlib140.dll, vcomp140.dll, vcruntime140.dll.
+Size = 13.7 MiB
+Category = 14
+URLSite = https://www.microsoft.com/en-us/download/details.aspx?id=53587
+URLDownload = https://download.microsoft.com/download/6/D/F/6DF3FF94-F7F9-4F0B-838C-A328D1A7D0EE/vc_redist.x86.exe
+SHA1 = 89f20df555625e1796a60bba0fbd2f6bbc627370
+CDPath = none
+SizeBytes = 14456872
+
+[Section.0407]
+License = Unbekannt
+Description = Benötigt für Visual Studio 2015 basierte Anwendungen. Beinhaltet: atl140.dll, concrt140.dll, mfc140.dll, mfc140u.dll, mfcm140.dll, mfcm140u.dll, msvcp140.dll, vcamp140.dll, vccorlib140.dll, vcomp140.dll, vcruntime140.dll.
+Size = 13,7 MiB
+
+[Section.0a]
+License = Desconocida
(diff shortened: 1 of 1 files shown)


vc2015run.txt
```diff
@@ -0,0 +1,65 @@
+[Section]
+Name = Microsoft Visual C++ 2015 Redistributable Update 3
+Version = 14.0.24212
+License = Unknown
+Description = Needed by applications compiled with Visual Studio 2015. Contains: atl140.dll, concrt140.dll, mfc140.dll, mfc140u.dll, mfcm140.dll, mfcm140u.dll, msvcp140.dll, vcamp140.dll, vccorlib140.dll, vcomp140.dll, vcruntime140.dll.
+Size = 13.7 MiB
```
learn-more: Is `Size` still needed if `SizeInBytes` is present?

oleg-dubinskiy: I think yes, because it always displays the size of the full downloaded file in the program info at the bottom of Rapps window unlike `SizeInBytes` which displays it only in download dialog and only during downloading. Will `Size` value be displayed if I'll remove it from the txt file?

learn-more: if `Size` is still used somewhere, that is a bug that should be fixed.

oleg-dubinskiy: I guess it's a bug in Rapps-db script itself, not from my side.

learn-more: Yes, it's a bug on the rapps side, not yours.
Sorry for the confusion :)

archeYR:
Is it really needed? MSVC 2017 libraries should be also compatible with applications compiled using MSVC 2015 compiler, if I'm not mistaken.

oleg-dubinskiy:
@archeYR As far as I remember, there are some older apps from 2015-2016, which don't know about MSVC 2017 libs, since that libs did not exist then, so they need exactly 2015 version installed.

archeYR:
@oleg-dubinskiy MSVC 2017 libraries are basically backward compatible with MSVC 2015 ones, it shouldn't be an issue (see https://docs.microsoft.com/en-us/cpp/windows/universal-crt-deployment). Do you know any example of software that fails with MSVC 2017 libraries, but work with MSVC 2015 ones?

ROCKNROLLKID:
Does MSVC 2019 also work on Windows XP? Apparently 2019 is also backward compatible with 2017/2015.

ROCKNROLLKID:
> @oleg-dubinskiy MSVC 2017 libraries are basically backward compatible with MSVC 2015 ones, it shouldn't be an issue (see https://docs.microsoft.com/en-us/cpp/windows/universal-crt-deployment). Do you know any example of software that fails with MSVC 2017 libraries, but work with MSVC 2015 ones?

Yes, most projects will be, however, .Net Core projects are not because project.json (2015) was changed to .csproj (2017). This is probably the only reason why you would need both 2015 and 2017 and why windows keep both 2015 and 2017 on your system.

On a related note, it seems 2019 is also backward compatible (minus the exception above) and should run on WinXp as well.

archeYR:
.NET Core redistributables are Windows 7+, so I guess it doesn't really matter here (at least for now). 2019 redistributables should work too (and the 2017 ones in RAPPS could be probably upgraded), but MSVC 2019 compiler doesn't support XP targetting, so you're still limited to applications compiled using MSVC 2017 or older. The only use case (in ROS) I could think of is a fallback solution, when an application for whatever reason fails in ROS using MSVC 2017 libs, but works using 2015 ones (although I guess it's unlikely to happen).

PR #64, opened 2019-06-21
Use general code for Arabic language

moceap:


HBelusca:
Is this needed because otherwise it doesn't display in RAPPS?

moceap:
0x0401 code only for ar_SA
0x0001 code for all ar

PR #65, opened 2019-07-24
[RAPPS-DB] Update 7-Zip to 19.00

shytzedaka:
I Have tested and Can Confirm that Version 19.00 Works on ReactOS
Screenshot (694)

Might Fix: https://jira.reactos.org/browse/CORE-16168
Since it uses a Different Download Link (The Original One used in the 7-Zip website)

### DIFF
--- a/7zip.txt
+++ b/7zip.txt
@@ -1,15 +1,15 @@
 [Section]
 Name = 7-Zip
-Version = 18.05
+Version = 19.00
 License = LGPL
 Description = A file archiving utility with support for 7zip, zip, tar, rar and many other archive formats.
 Size = 1.1 MiB
 Category = 12
 URLSite = http://www.7-zip.org/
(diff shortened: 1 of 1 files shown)


HBelusca:
Does it extract files correctly without making ROS crash?

shytzedaka:
> Does it extract files correctly without making ROS crash?

Yup. No issues at all. i Downloaded. Opened it. and Installed it and it works Just fine as the other one.

shytzedaka:
> Does it extract files correctly without making ROS crash?

i'll test that out

shytzedaka:
i am Currently uploading a video with a Putty log.

It **is** possible to Extract files. Just.. Not With the mouse. Drag and Drop Doesn't seem to work. 
Pressing the Extract button on the program and extracting works. 

Trying to Open a .zip file with the 7-ZIP program Opens up 7-Zip on the root folder. which means you have to Manually get to it. and then you can read the content of the file.

Youtube video:

PR #66, opened 2019-07-24
[RAPPS-DB] Update VLC to version 3.0.7.1

shytzedaka:
Tested out The Latest Version Available For VLC and it Works Good with ReactOS, So the Purpose of this PR is Just to Update VLC to the latest version because it works.
Screenshot (695)

Video Playback works. Fullscreen works (although toolbar still shows up)
Sound Playback is Unknown Due to not having the sound drivers. but i doubt that their result is any Different than Version 3.0.3 that is currently on RAPPS

### DIFF
--- a/vlc.txt
+++ b/vlc.txt
@@ -1,26 +1,26 @@
 [Section]
 Name = VLC media player
-Version = 3.0.3
+Version = 3.0.7.1
 License = GPL
 Description = A fully featured FOSS media player with built-in codecs.
-Size = 38.3 MiB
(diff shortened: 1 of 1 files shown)


shytzedaka:
@HBelusca Awaiting your Confirmation.

BenNottelling:
I've tested in both Vbox and vmware player. 
In both it performs the same, in vmware player sound works in both versions
(vbox sound is just horrid in general)
So, lgtm

PR #67, opened 2019-07-28
[RAPPS-DB] Merged Python 2 and 3 to Just 'Python'

shytzedaka:
I think it Would be a pretty good idea to Merge these two in to One. there is No new python Version to support ReactOS Since, from 3.6 it requires windows server 2008 and above. and 3.5 Fails to install. as a result the version is still the same.


HBelusca:
And what about people who want to use Python 2.x? This is something separate from Python 3.x.

shytzedaka:
> And what about people who want to use Python 2.x? This is something separate from Python 3.x.

Well. my question is. why would they use an older version? and they could always download it from the FTP.

What kind of Difference in terms of functionality does Python 2 and Python 3 have?

shytzedaka:
On 2nd thought the more i read about here:
https://wiki.python.org/moin/Python2orPython3
Makes me realize this might Not be a good idea then.
Apologies.

PR #71, opened 2019-08-14
[RAPPS-DB] Add ReadMe file

katahiromz:
Add ReadMe file (`README.ENG`). This is a copy of `$(REACTOS)/base/applications/rapps/README.ENG`.

PR #72, opened 2019-08-14
[RAPPS-DB] Created a readme

BenNottelling:
Follow up to https://github.com/reactos/rapps-db/pull/71
Based off the same file there, but removed some info that wasn't relevant to this database. This is a markdown readme
At some point it would be a good idea to talk about the kinds of downloads RAPPS can't handle, but I'm not 100% what kinds it has issues with

Use[ this link](https://github.com/reactos/rapps-db/blob/9b95e06218f32331ab643c9be8113f2f82fe5c3f/readme.md) to preview it

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,0 +1,82 @@
+For more information about RAPPS, please take a look at the [wiki](https://reactos.org/wiki/RAPPS) 
+
+ADDING PROGRAMS TO THE RAPPS DATABASE
+--------------------------------------------------
+
+Each program entry consists of a text file formatted with an INI-like syntax.
+
+They must be:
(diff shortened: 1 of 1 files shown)


katahiromz:
Rename it as `README.md`. Uppercase is the better.

BenNottelling:
> Rename it as `README.md`. Uppercase is the better.

I tried to rename it but git didn't track it as a change, so instead I did a forcepush. Seems the PR hasn't updated, but you can clearly see it here
https://github.com/BenNottelling/rapps-db/commit/f4dbeeefdca5f4382bce6b75ec5170bdcd748c00

katahiromz:
Once try another name.

BenNottelling:
I think github just hasn't noticed the change like git, but I'm assuming it would be fixed upon a merge

PR #78, opened 2019-08-14
Add AOL's Instant Messenger version 5.1 to RAPPS.

Xcallono:
Additionally, further fixing on Shy's attempt to implement my backgrounds into RAPPS. Hopefully this will be the last fix to that file because I'm pretty sure we're all sick of seeing the same file get patched like 6 times in the course of less than 12 hours.

### DIFF
--- a/aim.txt
+++ b/aim.txt
@@ -1,0 +1,12 @@
+[Section]
+ Name = AIM
+ Version = 5.1
+ License = Abandonware
+ Description = AOL Instant Messenger. Change host DNS to iwarg.ddns.net at port 5190 to restore functionality. Requires an AIM Phoenix account.
+ Size = 2.47 MB
+ Category = 16
+ URLSite = http://iwarg.ddns.net/
+ URLDownload = http://iwarg.ddns.net/aim/files/aim513036.exe
+ SHA1 = 0F6D27D02FA3ECA6ACBC4FC8F4D564B0D8D7AC76
+ CDPath = none
+ SizeBytes = 2598120
--- a/xbgpack1.txt
+++ b/xbgpack1.txt
@@ -1,12 +1,12 @@
 [Section]
-Name = Super Mario 64 Backgrounds!
-Version = 1
-License = CC0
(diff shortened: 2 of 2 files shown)


Xcallono:
I just got a report that the SHA1 hash was incorrect for my background pack, but this commit is supposed to fix it by downloading the original file the hash was based on, but from my website.

aim.txt
```diff
@@ -0,0 +1,12 @@
+[Section]
+ Name = AIM
+ Version = 5.1
+ License = Abandonware
+ Description = AOL Instant Messenger. Change host DNS to iwarg.ddns.net at port 5190 to restore functionality. Requires an AIM Phoenix account.
+ Size = 2.47 MB
+ Category = 16
+ URLSite = http://iwarg.ddns.net/
+ URLDownload = http://iwarg.ddns.net/aim/files/aim513036.exe
```
learn-more: Please don't use a `ddns.net` address for this.

Xcallono: What is the problem with a ddns.net address? Fill me in. This download is coming from another guy's website(virus checked and all, not that a virus on a ReactOS installation will do much at this stage.)

learn-more: ddns is a service that forwards a hostname to a dynamic ip,
which is usually used to connect one's private computer / synology to the internet, and reach it from the outside.
This means that the hosting most likely will not be reliable, and can result in reports that it's down at times.

Xcallono: So let me get this straight. You're asking me to change a link which, for every time I've ever used it, has been functional, by closing this merger that fixes another REALLY important issue(I have to close it because I screwed up the implementation of this merger by using master branch without thinking and now there's no way for me to fix this merger with Git Bash cleanly.) I mean, if it's really THAT bad of an issue, I will, but goodness me, this is embarrassing.

Xcallono:
Sorry about putting up a fuss about this, I should not have responded to you at 2 AM. I'll make some changes on my next PR when it's ready to solve this issue.

PR #80, opened 2019-08-20
Update inkscape 0.92.4

luzpaz:


HBelusca:
It works ok on ReactOS I suppose? (or at least the same as the previous version of inkscape?)

PR #84, opened 2019-10-03
Update npp.txt

marcinszynkowski:
Updated link for notepad++ download.
Right now the link provided is not available. It causes the rapps to show an error message about download error.
I've updated it to version available on the notepad++ official website.

### DIFF
--- a/npp.txt
+++ b/npp.txt
@@ -1,6 +1,6 @@
 [Section]
 Name = Notepad++
-Version = 6.9.2
+Version = 6.9
 License = GPL
 Description = A full featured, free text editor with many plug-ins.
 Size = 4.0 MiB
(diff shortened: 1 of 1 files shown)


learn-more:
Did you test this version in ReactOS ?

marcinszynkowski:
> Did you test this version in ReactOS ?

It's lower version than 6.9.2 that it was in the database - I assume that it worked, the available is 6.9.
But to be sure, I will test this version. I will update when finish.

marcinszynkowski:
rosnpp
It works, I've tested it.

JoachimHenze:
I think the version chosen by marcinszynkowski is ok.
Starting with v7.0 Notepad++ updater GUP.exe depends
on a libcurl.dll which is compiled for Vista+ (unnecessarily).

marcinszynkowski:
> You forgot to update version value.
> 
> ```diff
> -Version = 6.9.2
> +Version = 6.9
> ```

Thank you for spotting a mistake. 
I've fixed that :)

PR #85, opened 2019-10-05
[RAPPS-DB] Update CCleaner to v5.62

oleg-dubinskiy:
Current CCleaner 5.43 from Rapps fails to install:
ccsetup543

So there is a reason to update it to the newest 5.62 version, which installs and works correctly:
ccleaner

## Proposed changes

- Update version number, size of the installer in bytes and in megabytes, download url and SHA1;
- Additionally change the website url from piriform.com to ccleaner.com, because entering piriform.com from the browser now redirects to ccleaner.com anyway;
- Also add language-specific website urls for Deutsch, Spanish and Russian sections, to allow open the localized versions of program's website with different ReactOS localizations.

### DIFF
--- a/ccleaner.txt
+++ b/ccleaner.txt
@@ -1,23 +1,26 @@
 [Section]
 Name = CCleaner
-Version = 5.43
+Version = 5.62
 License = Freeware
 Description = A decent tool for cleaning temporary files and general computer maintenance.
-Size = 15.1 MiB
+Size = 24.2 MiB
 Category = 12
-URLSite = http://www.piriform.com/ccleaner
-URLDownload = https://download.ccleaner.com/ccsetup543.exe
-SHA1 = c8c687249db9a7383a5157eb3860091d6eb5b763
+URLSite = https://www.ccleaner.com/ccleaner
+URLDownload = https://download.ccleaner.com/ccsetup562.exe
+SHA1 = e5c348ec9ff7905cb7f09a61a8686cb1e32154de
+
 CDPath = none
-SizeBytes = 15838840
+SizeBytes = 25441808
 
 [Section.0407]
 Description = Sehr gutes Bereinigungswerkzeug für den PC.
-Size = 15,1 MiB
+Size = 24,2 MiB
+URLSite = https://www.ccleaner.com/de-de/ccleaner
 
 [Section.0a]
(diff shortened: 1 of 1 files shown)


JoachimHenze:
I just retested seconds ago: the old "CCleaner 5.43.6522" does install properly in ros 0.4.13-RC-7-g67211fa and its installer did not crash for me. Did you have Gecko installed when you tested Oleg?

I don't oppose updating the app, but want to mention pro forma:
- we still have JIRA tickets linked against the older version
- newer versions were recurrently reported of containing malware by different techsites
- and the new version is almost twice the size of our old

JoachimHenze:
Due to my scepticism regarding that app in general I uploaded the new 5.62 of Oleg with SHA1 e5c348ec9ff7905cb7f09a61a8686cb1e32154de to virustotal:
Just 1 of 62 scanners reported it, I assume that one could be a false alert
https://www.virustotal.com/gui/file/ce71e599cb5c07544f6db943ab918569aa707eac9151442be35ce5fe102ab56b/detection

oleg-dubinskiy:
> Did you have Gecko installed when you tested Oleg?

Yes, Gecko was installed each time when I tested it. Now I found out that it fails for me in fresh master only if to select Russian localization in the installer, what I did each time too. So looks like you are right. You probably tried to install it only with English language?
Btw, I already created the ticket for that earlier: https://jira.reactos.org/browse/CORE-16087
ccleaner543
ccleaner543rus2

JoachimHenze:
I always and only used ros configured to german localization when successing.
I never tried russian up to now. Thx for pointing out that difference.

learn-more:
> I just retested seconds ago: the old "CCleaner 5.43.6522" does install properly in ros 0.4.13-RC-7-g67211fa and its installer did not crash for me. Did you have Gecko installed when you tested Oleg?
> 
> I don't oppose updating the app, but want to mention pro forma:
> 
> * we still have JIRA tickets linked against the older version
> * newer versions were recurrently reported of containing malware by different techsites
> * and the new version is almost twice the size of our old

If you need, the old executable is available under:
http://web.archive.org/web/20190424042312/https://download.ccleaner.com/ccsetup543.exe
(For the jira tickets)

PR #86, opened 2019-11-02
[RAPPS-DB] ReactOS JPN Package 0.5

katahiromz:
JIRA issue: [CORE-12748](https://jira.reactos.org/browse/CORE-12748)

"ReactOS JPN Package" provides Japanese fonts and font substitute settings.

Source: https://github.com/katahiromz/ReactOS-JPN-Package
Binary: https://github.com/katahiromz/ReactOS-JPN-Package/releases/download/0.5/reactos-jpn-setup-0.5.exe

### DIFF
--- a/jpn-package.txt
+++ b/jpn-package.txt
@@ -1,0 +1,16 @@
+[Section]
+Name = ReactOS JPN Package
+Version = 0.5
+License = GPLv3
+Description = ReactOS JPN Package provides Japanese fonts and font substitute settings.
+Size = 4.8 MiB
+Category = 16
+URLSite = https://github.com/katahiromz/ReactOS-JPN-Package
+URLDownload = https://github.com/katahiromz/downloads/raw/master/RisohEditor-5.2.7.exe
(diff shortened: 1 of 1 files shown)


katahiromz:
successfully
Successfully in both of English ReactOS and Japanese ReactOS.

JoachimHenze:
I tried and I think the URL is fine. Unfortunately rapps can not download from git (at least from ros)
fixme:(dll/win32/wininet/cookie.c:988) secure not handled

You should double-check your txt from within rapps before committing.
I tried: It does not work for me neither with nor without that "?raw=true" at URLs end.

Contrary we do have some links to github.com that do still work, e.g:
https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/gs923w32.exe

Hope this helps in finding out how the URL needs to be modified.

JoachimHenze:
Downloading https://github.com/katahiromz/downloads/blob/master/reactos-jpn-setup-0.5.exe?raw=true from my browser in Windows does work though.

katahiromz:
I changed the link to one of the release page. Successful.
ok
ok2
ok3

PR #91, opened 2020-03-16
[RAPPS-DB] Add AssaultCube 1.2.0.2

oleg-dubinskiy:
assaultcube

AssaultCube is a free 3D first-person shooter game based on Cube engine.
Works fine in ReactOS, with Mesa3D from Rapps and with VBoxOGL renderer. Although with Mesa3D it has some minor issues: the map at the top right does not render properly (is invisible) and the game has lower FPS than with VBoxOGL.
I guess these are Mesa3D issues, not ReactOS, and on XP/2003 it will show same (or, at least, similar) result, although I did not test it yet.

### DIFF
--- a/assaultcube.txt
+++ b/assaultcube.txt
@@ -1,0 +1,22 @@
+[Section]
+Name = AssaultCube
+Version = 1.2.0.2
+License = Freeware
+Description = 3D first-person shooter based on Cube engine.
+Size = 43.3 MiB
+Category = 4
+URLSite = https://assault.cubers.net/
+URLDownload = https://github.com/assaultcube/AC/releases/download/v1.2.0.2/AssaultCube_v1.2.0.2.exe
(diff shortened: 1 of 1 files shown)


learn-more:
does it work without mesa?

oleg-dubinskiy:
Yes, and even better than with Mesa3D. I already said about this.

learn-more:
Then let me rephrase:
Does it work without Mesa and VboxOGL?

oleg-dubinskiy:
> Does it work without Mesa and VboxOGL?

Yes, it does, although with very low FPS:
VirtualBox_ReactOS_17_03_2020_21_14_07

But is it a mandatory requirement to add the games in Rapps? :thinking:

learn-more:
No, but it provides a better experience, as long as we don't have hints implemented that users should install something additionally.

PR #92, opened 2020-03-28
[RAPPS-DB] Add Sylpheed 3.7.0

oleg-dubinskiy:
sylpheed
Sylpheed is an open source lightweight e-mail client, which seems to be a good alternative to Mozilla Thunderbird, especially for old PCs.
Works great in ReactOS, unlike many other GTK+ apps, which hang at startup (because it uses an older GTK2 version, which works correctly in ROS). Also imports and reads the messages from my Gmail as well (after allowing the access and enabling POP3 in the Google account settings).
The only very minor issue is a bit slow drawing after closing program's settings/about dialog or other program's windows.


sylpheed.txt
```diff
@@ -0,0 +1,20 @@
+[Section]
+Name = Sylpheed
+Version = 3.7.0
+License = GPL
+Description = Lightweight and user-friendly e-mail client.
+Size = 7.85 MiB
```
learn-more: If `SizeBytes` is specified, this one is not needed anymore :)
Rapps will automatically format the `Size` according to system settings.

oleg-dubinskiy: Removed.
But I wonder: if `Size` and `SizeBytes` are different, will it display the same correct `Size` value? Or it will be just converted from B into MB? E. g. 8.23 MB instead of 7.85.

sylpheed.txt
```diff
@@ -0,0 +1,19 @@
...
+Category = 5
+URLSite = https://sylpheed.sraoss.jp/en/
+URLDownload = https://sylpheed.sraoss.jp/sylpheed/win32/Sylpheed-3.7_setup.exe
+SHA1 = 0518235bbef0cda374e1c8002eb4f65f8f2f8e6c
+CDPath = none
+SizeBytes = 8233467
+
+[Section.0419]
+Description = Легковесный и дружелюбный к пользователю почтовый клиент.
+Size = 7,85 МиБ
```
learn-more:
```suggestion
```

sylpheed.txt
```diff
@@ -0,0 +1,19 @@
...
+
+[Section.0422]
+Description = Легкий і дружній до користувача поштовий клієнт.
+Size = 7,85 МіБ
```
learn-more:
```suggestion
```

sylpheed.txt
oleg-dubinskiy: Committed, thanks.
Forgot about that.

PR #94, opened 2020-05-05
[ADDITION] JET40 and MDAC28SP1

julenuri:
I want to add those two Microsoft libraries.
* JET 40 - I added the only JET compatible afaik (I tried the 2003 server and the WinMe ones).
* MDAC28SP1 - I added, you have to put this instalator in Windows 2000 mode to run it, two times and it installs.

### DIFF
--- a/jet40.txt
+++ b/jet40.txt
@@ -2,8 +2,7 @@
 Name = Jet 4.0 Service Pack 8 (SP8) (KB829558) for 9x/NT
 Version = 4.0.7328
 License = Unknown
-Description = This installation includes Microsoft products that incorporate Microsoft Jet database engine version 4.0. 
-Size = 3.7 MiB
+Description = This installation includes Microsoft products that incorporate Microsoft Jet database engine version 4.0.
 Category = 14
 URLSite = https://www.microsoft.com/en-US/download/details.aspx?id=7151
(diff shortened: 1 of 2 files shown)


mdac28.txt
```diff
@@ -0,0 +1,16 @@
+[Section]
+Name =  Microsoft Data Access Components (MDAC) 2.8 SP1
+Version = 1.0
+License = Unknown
+Description = MDAC 2.8 SP1 contains core Data Access components such as the Microsoft SQL Server™ OLE DB provider and ODBC driver.
+Size = 5.8 MiB
```
HBelusca, line 6: Not needed, since you specify the SizeBytes entry below.

jet40.txt
```diff
@@ -0,0 +1,18 @@
+[Section]
+Name = Jet 4.0 Service Pack 8 (SP8) (KB829558) for 9x/NT
+Version = 4.0.7328
+License = Unknown
+Description = This installation includes Microsoft products that incorporate Microsoft Jet database engine version 4.0. 
+Size = 3.7 MiB
```
HBelusca, line 6: Not needed since you specify the SizeBytes entry below.

julenuri: Sorry for both

JoachimHenze:
julenuri, What application does rely on those libs and can be started successfully on ros after installing Jet40 / MDac 2.8sp1?

PR #95, opened 2020-05-12
[ADDITION] Add Zeth.mstyles

Mas4hmad:
Add variation for visual styles

### Preview:
Untitled

### DIFF
--- a/zeth.txt
+++ b/zeth.txt
@@ -3,12 +3,11 @@ Name = Zeth
 Version = 0.0.1
 License = Freeware
 Description = Zeth Visual Style for ReactOS
-Size = 116 KiB
 Category = 15
 URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
-SHA1 = D06F5C1ADB15FB892F07583133D02E23AC3575C7
+SHA1 = F905698A0AA0CBC3F5AC1F3C1B9019444608DFB8
 CDPath = none
-SizeBytes = 119.658 bytes
+SizeBytes = 618496
 
 
 
--- a/zeth.txt
+++ b/zeth.txt
@@ -1,0 +1,14 @@
+[Section]
+Name = Zeth
+Version = 0.0.1
+License = Freeware
+Description = Zeth Visual Style for ReactOS
+Size = 116 KiB
+Category = 15
+URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
+SHA1 = D06F5C1ADB15FB892F07583133D02E23AC3575C7
+CDPath = none
+SizeBytes = 119.658 bytes
+
+
+
--- a/zeth.txt
+++ b/zeth.txt
@@ -2,12 +2,12 @@
 Name = Zeth
 Version = 0.0.1
 License = Freeware
-Description = Zeth Visual Style for ReactOS
+Description = Zeth Visual Style for ReactOS. This is the executable files, so you can extract this files to C:\ReactOS\Resources\Themes\Zeth executionally.
 Category = 15
-URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
-SHA1 = F905698A0AA0CBC3F5AC1F3C1B9019444608DFB8
+URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.exe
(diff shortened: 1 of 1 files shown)


Mas4hmad:
i use selawik font with different version because from ReactOS is shifted up little bit. i'll do upload with separate PR if you please. Thanks

zeth.txt
```diff
@@ -0,0 +1,14 @@
...
+Name = Zeth
+Version = 0.0.1
+License = Freeware
+Description = Zeth Visual Style for ReactOS
+Size = 116 KiB
+Category = 15
+URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
+SHA1 = D06F5C1ADB15FB892F07583133D02E23AC3575C7
+CDPath = none
+SizeBytes = 119.658 bytes
```
learn-more, line 11:
```suggestion
SizeBytes = 119658
```

zeth.txt
```diff
@@ -0,0 +1,14 @@
+[Section]
+Name = Zeth
+Version = 0.0.1
+License = Freeware
+Description = Zeth Visual Style for ReactOS
+Size = 116 KiB
```
learn-more, line 6:
```suggestion
```

zeth.txt
```diff
@@ -0,0 +1,14 @@
...
+Category = 15
+URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
+SHA1 = D06F5C1ADB15FB892F07583133D02E23AC3575C7
```
learn-more, line 9: Are you sure this is the correct hash?

zeth.txt
learn-more: This is not needed anymore

zeth.txt
learn-more: This should just be a number, not any formatting

zeth.txt
```diff
@@ -0,0 +1,14 @@
+[Section]
+Name = Zeth
+Version = 0.0.1
+License = Freeware
+Description = Zeth Visual Style for ReactOS
+Size = 116 KiB
+Category = 15
+URLDownload = https://github.com/Mas4hmad/ms-visual-styles/raw/master/ZETH.msstyles
```
learn-more, line 8: Does ReactOS install it correctly like this?

learn-more:
To test this, disable the option 'Update the list of accessible programs at start,
and drop the file in the local rapps dir (Documents and settings\username\rapps I think)

zeth.txt
Mas4hmad: oh, nope. i input that SHA1 before commit into repo. i don't know if it SHA1 is different. will do correct

zeth.txt
Mas4hmad: that's not installer, just copy it manually into theme folder.

zeth.txt
Mas4hmad: thanks, will do correctly

Mas4hmad:
> 
> 
> To test this, ... .

i have test the text file download and it doing correctly. thanks for advising. wix fix this soon

zeth.txt
learn-more: In that case it would be a good idea to either create an installer, or add instructions in the description?

Mas4hmad: for my first step, i would prefer add instruction. instead, create an installer is less possible regarding my lack of C language, i will make 'create an installer' if i ready.

petr-akhlamov: I usual make installer via 7zip SFX Builder.
https://sourceforge.net/projects/s-zipsfxbuilder/
You can open my sfx (*.exe) in it and learn how to make own.
https://github.com/petr-akhlamov/ros-repo/raw/master/lunabeta.exe

learn-more: All those installers ask the user where to install (and provide a correct default),
it's better to use f.e. winrar that allows to specify the target directory....

Mas4hmad: thanks for the hint, will do update.
thank you so much before

Mas4hmad:
> 
> 
> This url does not work for me...

it's weird. i have given correct url, but it can't download. but if i click link in description, it's download flawlessly.

Mas4hmad:
the problem solved. i was use wrong OS that using Rapps (it's Windows XP!). Should be fixed now in ReactOS.
thanks

PR #98, opened 2020-05-21
Added SDelete from Sysinternals Suite

zbalkan:
Includes both x86 and x64 versions in a zip file. Requires a zip extractor.

### DIFF
--- a/sdelete.txt
+++ b/sdelete.txt
@@ -1,0 +1,20 @@
+[Section]
+Name = SDelete
+Version = 2.02
+License = Sysinternals Software License Terms
+Description = SDelete is a free command line tool that you can use to delete files securely so that they cannot be recovered anymore. You will need a ZIP decompression program to install it.
+Size = 221 KiB
+Category = 12
+URLSite = https://docs.microsoft.com/en-us/sysinternals/downloads/sdelete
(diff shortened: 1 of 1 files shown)


sdelete.txt
```diff
@@ -0,0 +1,20 @@
+[Section]
+Name = SDelete
+Version = 2.02
+License = Sysinternals Software License Terms
+Description = SDelete is a free command line tool that you can use to delete files securely so that they cannot be recovered anymore. You will need a ZIP decompression program to install it.
+Size = 222 KiB
```
HBelusca, line 6: You can get rid of this "Size" field, which is deprecated; the SizeInBytes field from below is better.

zbalkan: I kept that because wiki has this warning about Size section. 

"DEPRECATED: The value should be kept for backward compatibility."
As you say I can get rid of it, then I'll do it asap.

HBelusca: It would be backward compatibility with older versions of RAPPS that you get only by downloading old versions of ReactOS.

learn-more: Thanks, I fixed the wiki article

zbalkan: Anything else?

PR #103, opened 2020-08-13
Fixes some HTTPS URLS

KyleKatarn31:
7zip, Abiword site URL should use HTTPS, not HTTP


learn-more:
Did you test the downloads in reactos?

KyleKatarn31:
Since HTTPS URLS are legit and seems some other products are already using HTTPS url for download, what could be the issue ? Any specific precaution or guidance ?

learn-more:
> Since HTTPS URLS are legit and seems some other products are already using HTTPS url for download, what could be the issue ? Any specific precaution or guidance ?

Our ssl support is not on the same level as for win10.

KyleKatarn31:
Any specific procedure to test a local RAPPS DB ?

Swyter:
The files are check-summed with SHA-1. So validation is not a big issue here.
The only thing you gain with TLS in rapps is that people won't know the exact file you are downloading.

KyleKatarn31:
My suggestion is more on site URL, not necessarily on download url (see suggested changes)

learn-more:
> My suggestion is more on site URL, not necessarily on download url (see suggested changes)

and yet you change them.
Either check that they work in reactos, or revert it.

PR #104, opened 2020-08-13
Update CCleaner from 5.62 to 5.70

KyleKatarn31:
Update CCleaner from 5.62 to 5.70

### DIFF
--- a/ccleaner.txt
+++ b/ccleaner.txt
@@ -1,16 +1,13 @@
 [Section]
 Name = CCleaner
-Version = 5.62
+Version = 5.70
 License = Freeware
 Description = A decent tool for cleaning temporary files and general computer maintenance.
 Size = 24.2 MiB
 Category = 12
 URLSite = https://www.ccleaner.com/ccleaner
-URLDownload = https://download.ccleaner.com/ccsetup562.exe
-SHA1 = e5c348ec9ff7905cb7f09a61a8686cb1e32154de
-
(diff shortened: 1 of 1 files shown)


oleg-dubinskiy:
Does it install and work correctly for you in ReactOS? If yes, can you provide a screenshot of that please?
I had the problem with the latest versions of CCleaner in nightly ROS revisions not so long time ago. That was most of all a regression. It didn't start that time at all for me after installation.

KyleKatarn31:
Do you mean full feature from CCleaner needs to be functionnal ? Was it the case on 5.62 ?

KyleKatarn31:
My mistake, latest version for "xp-vista" is 5.64, not 5.70

oleg-dubinskiy:
Yes, the all features of free version, or nearly all (minor bugs are not particularly significant).
Yep, that happened for me with 5.62.

oleg-dubinskiy:
Ah no, doesn't work for me too. 5.70 indeed requires Windows 7+. :frowning_face: 
ccleaner_fail

oleg-dubinskiy:
Hmm, but 5.64 installs and works for me without major bugs. :smiley: 
ccleaner_5 64

KyleKatarn31:
5.64 installs and work fine with reactos-bootcd-0.4.15-dev-511-g90456e5-x86-gcc-lin-dbg

learn-more:
Stale

PR #118, opened 2020-09-17
Added an entry for Dave Gnukem

qbancoffee:
Dave Gnukem is a retro-style 2D scrolling platform shooter similar to, and inspired by, Duke Nukem 1
The source and assets are licensed under MIT GPL and CC
https://github.com/davidjoffe/dave_gnukem

I tested the download and install with RAPPS and it worked.
This game installed and ran perfectly even with sound using ReactOS 0.4.13

The creator of the game made some comments in his repo about making changes to the game so that it
could run under ReactOS.

This is the latest version


qbancoffee:
Screenshot from 2020-09-16 21-23-04
Screenshot from 2020-09-16 21-23-48
Screenshot from 2020-09-16 21-24-57

davegnukem.txt
```diff
@@ -0,0 +1,12 @@
+[Section]
+Name = Dave Gnukem
+Version = 1.01
+License = MIT GPL CC
+LicenseInfo = 1
+Description =  Dave Gnukem is a retro-style 2D scrolling platform shooter. It is inspired by and similar to Duke Nukem 1.
+Category = 4
+URLSite = https://github.com/davidjoffe/dave_gnukem
+URLDownload = https://sourceforge.net/projects/gnukem/files/gnukem/Dave_Gnukem_Setup_1.0.1.exe
```
learn-more, line 9: Did you test this url inside reactos?
I doubt that it will actually download the binary.

qbancoffee: Yes, I just took a screenshot of it working., check it out.

Screenshot from 2020-09-17 08-29-08

learn-more (approved):

davidjoffe:
For reasons unclear to me that v1.0.1 link, although it downloads the binary to the RAPPS download folder, it says 'fails integrity check' now (SHA1 and filesize are different) ... I'm guessing maybe I replaced the 1.0.1 exe on sourceforge at some stage possibly  ... (though there's a much newer version 1.0.3.2 version which I'd like to try include instead, I'm just doing some testing and v1.0.3.2 gives a runtime error about 'The procedure entry point ucrtbase.terminate could not be located in the dynamic link library api-ms-win-crt-runtime-l1-1-0.dll' so for now maybe best to stick to v1.0.1) ... The v1.0.1 binary seems to run if one runs it from the RAPPS download folder manualyl - I just tested it - but that doesn't help for 99% of normal users of course.) I'll try push a fix for v1.0.1 so long.

PR #122, opened 2020-10-02
Add Notepad2e

chirsz-ever:


HBelusca (changes requested):
Can you show us an example that proves that this application works fine in ReactOS?

chirsz-ever:
VirtualBox_ReactOS-0 4-2010022355

notepad2e.txt
```diff
@@ -0,0 +1,11 @@
+[Section]
+Name = Notepad2e
+Version = 4.2.25
+License = 3-clause BSD
+Description = Notepad 2e is a fork of Notepad2 with word highlighting, simultaneous editing, math evaluation, un/grep, UAC elevation, complete regexps (PCRE), Lua lexers, DPI awareness and more 
+Size = 980 KiB
```
learn-more, line 6: Size is not needed when SizeBytes is used.

chirsz-ever: This should be mentioned in [the wiki](https://reactos.org/wiki/RAPPS).

learn-more:
Thanks!

PR #126, opened 2020-10-06
add Winfile to rapps

qbancoffee:
This is a slightly modified version of winfile so that it can be compiled
by mingw for Windows Server 2003 SP2. There is a small hack that allows
it to run on ReactOS.

[Click here to see a video of installation via rapps and WinFile running.](https://youtu.be/2FXLrNfeLC0)

Screenshot from 2020-10-06 13-30-34
Screenshot from 2020-10-06 13-30-45

GNUmakefile was modified so that it can cross compiled on linux for
Windows.
This release was compiled on Ubuntu 18.04 with

i686-w64-mingw32-g++ (GCC) 7.3-win32 20180312
i686-w64-mingw32-gcc (GCC) 7.3-win32 20180312

The installer was created using NSIS v2.51 on Ubuntu 18.04 under wine.
The winfile.exe was statically linked so it's a stand alone binary.


qbancoffee:
A check is failing but I can't seem to figure out what it is.......

binarymaster:
UCS-2 Big Endian encoding...

qbancoffee:
Finally! somehow git was re-encoding the file...... I pushed my changes from another terminal and it worked.

binarymaster (changes requested):
Looking at the [file mirror](https://github.com/qbancoffee/winfile/releases/tag/v1.0) you provided as download link, it states "There is a small hack that allows it to run on ReactOS".

The purpose of RAPPS database is not to include every program that runs on ReactOS, but to include programs that work there without any modifications.

You should first investigate the problem why it does not run or install on ReactOS without hack, and once this is addressed, you have to provide download link for the program without modifications or hacks.

qbancoffee:
While I agree with you that the aim is to run programs unmodified,  the version I modified isn't even the original version of WinFile and I'm not even sure that the original version is available for download(wasn't it Win16?).

If you read the Mircosoft WinFile repo it  says that the source code Microsoft provides has personal edits and was modified to compile on Visual Studio and on supported versions of Windows including Windows 10

If I remember correctly(It's a bit hazy), all I did was add statements like "#if (_WIN32_WINNT >= NTDDI_WS03SP2)" 
so that the produced binaries that could run  on Windows server 2003 SP2 once again. I also created a Makefile so that it could compile with gcc and the link to the binary is one that I compiled on my machine.
I think there was also an incompatibility with "GetProcAddress" or something. 

It certainly isn't a replacement for the existing file manager but rather a piece of history that one could install and play with. It's a curiosity and really isn't any different than what Microsoft is currently doing with it's modified version of WinFile.
 At the moment you can install WinFile from the app store and run it on your Windows 10 machine. 

If it's any comfort this version runs on Windows XP as well.

qbancoffee:
These Win32 functions are in WinFile but the minimum requirement is Windows server 2008 for these.

LocaleNameToLCID
GetLocaleInfoEx

I added this to 
https://jira.reactos.org/browse/CORE-14376

learn-more (approved):

Extravert-ir (approved):

tkreuzer (approved):

PR #120, opened 2020-09-24
Add Calibre 2 to the software available list

julenuri:
Adding Calibre 2.0.0 to the software available list.
It is interesting to have it as reader and potentially connect them with other devices for reading PDF, MOBI, EPUB and DOC.

PR #145, opened 2021-08-16
[RAPPS-DB] Add Calibre 2.0.0 to the Rapps-db in x86/AMD64

julenuri:
Fix of the #120. Now I know how to use Git. 

Adding Calibre 2.0.0 to the software available list.

It is interesting to have it as reader and potentially connect them with other devices for reading PDF, MOBI, EPUB and DOC.
Images that probe that works:

Snapshot x86

### DIFF
--- a/calibre2.txt
+++ b/calibre2.txt
@@ -1,0 +1,18 @@
+[Section]
+Name = Calibre
+Version = 2.0.0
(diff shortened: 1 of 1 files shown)


learn-more:
Thanks!

PR #147, opened 2021-09-06
Documentation

d1nuc0m:
I am sorry for using a pull request in this way, but it seems like Issues are not enabled (as a sidenote, can I ask why?)

I have made some minor improvements to Readme formatting, and would like to discuss some doubts about the parameters (I've written them in bold, with __ to distinguish them from readme content).

Another question, not included. In some files, like 7-zip one, there is a [Section.amd64], as the x64 is not fully working yet, is this a experimental feature? Has this parameter been enabled yet or it was just an example)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
+# RAPPS DATABASE 
 For more information about RAPPS, please take a look at the [wiki](https://reactos.org/wiki/RAPPS) 
 
-ADDING PROGRAMS TO THE RAPPS DATABASE
---------------------------------------------------
+## ADDING PROGRAMS TO THE RAPPS DATABASE
 
 Each program entry consists of a text file formatted with an INI-like syntax.
 
@@ -9,15 +9,16 @@ They must be:
 Encoded in `UTF-16 LE (Little Endian)` or characters out of the ANSI range will display broken. 
 **Note:** some editors like Notepad++ call this format `UCS-2 Little Endian.`
 
+### Sections
 
 Each `[Section]` is language-independent and individual, you can override the URL to a source program or any other field by adding a language-specific `[Section.]`, followed by the language code.
 
 **Note:** You can find a complete listing of LCIDs and language names on [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/dd318693%28v=vs.85%29.aspx), includes neutral codes:
+
+**__Link is "broken", it works but redirects [here](https://docs.microsoft.com/en-us/windows/win32/intl/language-identifier-constants-and-strings), that does not show the codes anymore, as it has been deprecated as system, this information is now [here at page 31 and next](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-LCID/%5bMS-LCID%5d.pdf)__**
      
 RAPPS also accepts neutral language codes, meaning that you can do things like this:
-
  
-
      ; Default English fallback, used if everything else fails.
       [Section]
       Name = Name in English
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -33,12 +34,50 @@ RAPPS also accepts neutral language codes, meaning that you can do things like t
 You can also define an entry without English fallback to make it visible to certain users only.
 For instance; software from 1C, which is mostly for Russian speakers and unusable for anyone else.
 
+**__ There are some user that can speak multiple languages, for example an English-native user (with the OS set to English)  could also speak German (and use German-only software, but in case couldn't see it listed). Maybe there could be a way for the user to set which languages he would like to be listes? __**
```
learn-more, line 37: sure, feel free to submit a PR.

README.md
```diff
@@ -33,12 +34,50 @@ RAPPS also accepts neutral language codes, meaning that you can do things like t
...
+This is the version of the program.  **__ As it is used to check for updates, which decimal separator it uses? Comma or dot? Does it use the OS language one? Or is it fixed? __**
+
+### License
+The license that the program is based on. Typical options are : Trial, Demo, Freeware, Open-Source. Or more specifically, GPL, MIT and so on. **__ What happens if it used together with LicenseInfo? Does it also support links? __**
+
+### Description
+The description of the program giving some additional background information about what it does.
+
+### SizeBytes
+The actual size of the download in bytes. Used to display size value in the application info as well as while downloading.
```
learn-more, line 77: I don't see the added value in duplicating the information from the wiki, that means updating even more documentation (which is already hardly ever done, so surely adding more work is not gonna help)

README.md
```diff
@@ -57,4 +96,45 @@ List of valid categories:
  15. Themes
  16. Other
 
-The official list of downloadable programs is kept on a public ReactOS server and synced every time RAPPS is launched for the first time.
+### URLSite
+Main web site where the program can be found. **__ Is HTTPS mandatory or only advised? If the link is in HTTP, can it try to switch automatically to HTTPS if avaliable? __**
```
learn-more, line 100: not mandatory at all.

README.md
```diff
@@ -57,4 +96,45 @@ List of valid categories:
...
+
+### URLDownload
+Direct link to the installer/program. **__ Is HTTPS mandatory or only advised? If the link is in HTTP, can it try to switch automatically to HTTPS if avaliable? __**
+
+### ScreenshotN
+Screenshot URL link. (N = 1, 2, 3 ...)
+
+Currently, only first Screenshot is used. Ensure the URL is a direct link to the image, not a page containing the image (e.g. Imgur links). It is also preferred to include the screenshot taken on ReactOS, not on Windows.  **__ Does it still support only one screenshot? Have you thought about creating a repository for screenshots? __**
```
learn-more, line 108: only one screenshot, yes.
We did think about creating something for that, but there are enough free image hosts (like imgur) that specialize in just this.
We are building an OS, we don't really need to also replicate every website and service out there.

README.md
```diff
@@ -57,4 +96,45 @@ List of valid categories:
...
+
+### Icon
+Icon filename in icons folder (with .ico extension) **__ Any specification on size etc? __**
```
learn-more, line 111: Don't think so, but a good starting point would be looking at where they are used, and the existing ones.

README.md
```diff
@@ -57,4 +96,45 @@ List of valid categories:
...
+* Languages
+* LicenseInfo
+
+### Languages
+This field is used to inform the user whether the app is available in their language.
+
+You should place all the language codes app supports separated by | there. 
+Example
+     Languages=0C09|0813|0422
+**__ How are these code formatted? Are the same used in Sections? __**
```
learn-more, line 129: see the ones already doing this....

README.md
```diff
@@ -57,4 +96,45 @@ List of valid categories:
...
+### LicenseInfo
+This is a field in the DB which, when present, changes the way License field in the info is shown.
+
+Application license types correspond to a number:
+
+    1 - "Open Source"
+    2 - "Freeware"
+    3 - "Demo/Trial".
+   
+**__ How does this change the way the License is shown? __**
```
learn-more, line 140: Try it?

README.md
d1nuc0m: I don't want to duplicate the wiki, I want to update it. The problem is that I have some questions  to ask and it is not possible to open issues on this repository (otherwise I would just have opened one). 

All the parts marked as `**__ {text} __**`, are the questions

README.md
d1nuc0m: Ok , thank you

README.md
d1nuc0m: Ok, apparently same language codes. Are users advised to add the Languages parameter or to avoid it because it is experimental?

README.md
d1nuc0m: Ok, it was just to not have to depend on external services. Is there a way to document screenshot's copyright, apart from editing the image? (Something like Wikipedia, that has templates for screenshots of libre/copyrighted software etc.)

d1nuc0m:
Thank you for the help.
To sum up the questions (and add a few ones). Striken-through have been answered:
* ~~Sections: There are some user that can speak multiple languages, for example an English-native user (with the OS set to English) could also speak German (and use German-only software, but in case couldn't see it listed). Maybe there could be a way for the user to set which languages he would like to be listes?~~
* Version:  As it is used to check for updates, which decimal separator it uses? Comma or dot? Does it use the OS language one? Or is it fixed?
* License: What happens if it used together with LicenseInfo? Does it also support links?
* ~~URLSite: Is HTTPS mandatory or only advised? If the link is in HTTP, can it try to switch automatically to HTTPS if avaliable?~~
*  CDPath: it is undocumented on the wiki and in every file here is set to none (even if it is not mandatory). Has it been deprecated? Or is it in use and what should it do?
* SHA1: is it used to verify the file or is it here as a proposal? Are also other hashes supported?
*  Architecture: In some files, like 7-zip one, there is a [Section.amd64], as the x64 port is not fully working yet, is this a experimental feature? Has this parameter been enabled yet or it was just an example?
* ~~ScreenshotN: Does it still support only one screenshot? Have you thought about creating a repository for screenshots?~~
* ~~Icon: Any specification on size etc?~~ 
* Languages: ~~How are these code formatted? Are the same used in Sections?~~ Are users advised to add the Languages parameter or to avoid it because it is experimental?
* LicenseInfo: ~~How does this change the way the License is shown?~~

learn-more:
A PR is not really a good place to learn how to do your own research.
* We have a wiki: https://reactos.org/wiki/
* We have a bugtracker: https://jira.reactos.org/secure/Dashboard.jspa
* We have a discussion platform: https://chat.reactos.org/
* We have multiple mailing list: https://reactos.org/mailing-lists/

Considering there are more than enough ways to gather the information you are asking in your PR, there really is no need to add yet another way (github discussions / issues)

PR #148, opened 2021-09-10
[RAPPS-DB] Add Audacious 3.5.2

oleg-dubinskiy:
Add Audacious audio player 3.5.2 (old version) in Rapps.
Although it has some minor issues:
- sometimes small graphical glitches may appear in skinned mode with some skins;
- a text is not displayed correctly in URL bar when try to open an Internet stream from a web-link.

But this version actually works in ReactOS (plays local media files and (web) playlists fine at least), unlike all newer ones, those use the newer versions of GTK and just hang at start (known and already reported issue).

Additionally, add the icon extracted from the app's source. The app's license (BSD 2-Clause) seems to allow it without any restrictions.

Here are some screenshots:

audacious-3 5 2-2
audacious-3 5 2

Added links to them as well. :slightly_smiling_face:


audacious.txt
```diff
@@ -0,0 +1,24 @@
+[Section]
+Name = Audacious
+Version = 3.5.2
+License = BSD
+Description = An open source audio player inspired by Winamp/XMMS.
+Size = 18.4 MiB
```
learn-more, line 6: Size is not needed anymore

oleg-dubinskiy: OK, removed.

learn-more:
Thanks!

PR #149, opened 2021-09-23
[RAPPS-DB] 7-Zip - Update broken link, add icon

QrchackOfficial:
Just a little housekeeping. 32x32 ICO extracted from 7zFM.exe using Resource Hacker.

image


JoachimHenze:
Just for the record: What you do here is to replace "one non-working URL" by "another non-working URL".
It will work from Windows, but it will not work from ReactOS yet, due to a bug in our crypto-libs https://jira.reactos.org/browse/CORE-16741
Can you provide any non- "www.7-zip.org" URL? E.g. from archive.org?

QrchackOfficial:
www.7-zip.org also seems to serve the same file over HTTP by just rewriting the URL, if needed

JoachimHenze (approved):
If you tested it from within ros, then it should be fine.
Not so long ago 7-zip.org did strictly rely on that ECDSA_P256 algo. Maybe they improved their backward compatibility recently then.

QrchackOfficial:
Connecting to 7-zip.org, I get `ECDHE-RSA-AES128-SHA` over TLSv1.0/1.1, and ` ECDHE-RSA-AES128-GCM-SHA256` over TLSv1.2, confirmed using openssl just now.

PR #150, opened 2021-09-23
Update MyPal URL

Werru2955:
The Mypal binaries have been removed from the authors GitHub, and mypal has ceased development. However, I found an old archive of it on the wayback machine, and have made a mirror of it. This change switches the (now dead) link to my mirror.


mypal.txt
```diff
@@ -2,12 +2,11 @@
 Name = Mypal
 Version = 27.9.4
 License = MPL/GPL/LGPL
-Description = Mypal is a current and maintained Windows XP web browser.
+Description = Mypal is a current but abandoned Windows XP web browser.
 Size = 30.0 MiB
 Category = 5
-URLSite = https://github.com/Feodor2/Mypal
-URLDownload = https://github.com/Feodor2/Mypal27old/releases/download/27.9.4/mypal-27.9.4.win32.installer.exe
-SHA1 = 185ad6bd279d532789a8e890773bc95239273ec4
```
learn-more, line 10: Why did you remove the hash?

Werru2955: I didn't mean to. The SHA1 is still the same.

JoachimHenze:
Just for the record: What you do here is to replace "one non-working URL" by "another non-working URL".
It will work from Windows, but it will not work from ReactOS yet, due to a bug in our crypto-libs https://jira.reactos.org/browse/CORE-16741
Can you provide any non-github URL? E.g. from archive.org?

QrchackOfficial:
Couldn't get it to work on ReactOS 0.4.15-x86-dev, build 20210922-0.4.15-dev-3182-g7b62228.GNU_8.4.0:

image

Opening the URL in Wine Internet Explorer shows this prompt:

image

`wget` from "GNU utilities for Win32" reports `Unable to establish SSL connection`:

image

Firefox 48.0.2 downloads the file just fine

QrchackOfficial:
GitHub's certificate is from `DigiCert High Assurance EV Root CA`, which is a trusted root CA in Wine IE by default (verified that).

I get `TLS_AES_128_GCM_SHA256` cipher for TLS 1.3 and `ECDHE-ECDSA-AES128-GCM-SHA256` for TLS 1.2. TLS 1.1 throws an alert and won't connect.

Werru2955:
> 
> 
> Just for the record: What you do here is to replace "one non-working URL" by "another non-working URL".
> It will work from Windows, but it will not work from ReactOS yet, due to a bug in our crypto-libs https://jira.reactos.org/browse/CORE-16741
> Can you provide any non-github URL? E.g. from archive.org?

Got it, I'll change it to "https://archive.org/download/mypal-27.9.4.win32.installer/mypal-27.9.4.win32.installer.exe"

PR #159, opened 2021-11-19
Add otya128's WineVDM installer entry

peterooch:
Currently hosted on the repository containing the install script.
Can change it if something can be improved.

This PR is made with the reactos/reactos#4113 in mind as a follow-up to have a comfortable way to install otya128's WineVDM.
JIRA issue: [CORE-17852](https://jira.reactos.org/browse/CORE-17852)
This version is a GCC compiled version that includes all the required binaries.

The script itself is hosted in [peterooch/otvdm_installer](https://github.com/peterooch/otvdm_installer)

### DIFF
--- a/winevdm.txt
+++ b/winevdm.txt
@@ -1,6 +1,6 @@
 [Section]
 Name = WineVDM
-Description = Enables support for WOW16 applications. Installs in Program Files directory.
+Description = Enables support for WOW16 applications. Installs in the "Program Files" directory.
 Category = 14
 URLDownload = https://github.com/peterooch/winevdm_installer/releases/download/0.8.1-gcc-v2/winevdm_setup.exe
 SizeBytes = 3304645
--- a/winevdm.txt
+++ b/winevdm.txt
@@ -1,10 +1,10 @@
 [Section]
-Name = otvdm
-Description = Enables support for wow16 applications
+Name = WineVDM
+Description = Enables support for WOW16 applications. Installs in Program Files directory.
 Category = 14
-URLDownload = https://github.com/peterooch/otvdm_installer/releases/download/0.8.1-gcc/otvdm_setup.exe
-SizeBytes = 3304703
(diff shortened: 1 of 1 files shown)


winevdm.txt
```diff
@@ -0,0 +1,10 @@
+[Section]
+Name = otvdm
```
binarymaster, line 2: It would be better to name it "otya128's WineVDM" (see the chat for details)

winevdm.txt
```diff
@@ -0,0 +1,10 @@
...
+Description = Enables support for wow16 applications
```
binarymaster, line 3:
```suggestion
Description = Enables support for WOW16 applications.
```

winevdm.txt
```diff
@@ -0,0 +1,10 @@
...
+Category = 14
+URLDownload = https://github.com/peterooch/otvdm_installer/releases/download/0.8.1-gcc/otvdm_setup.exe
```
binarymaster, line 5: Would be nice to rename repository and installer names accordingly

binarymaster, line 5: Would be nice to rename repository and installer names accordingly.

winevdm.txt
oleg-dubinskiy: Or maybe better just "WineVDM". As I know, usually we don't specify developer/company names in the app's names.

winevdm.txt
oleg-dubinskiy, line 3:
```suggestion
Description = Enables support for WOW16 applications. Extract in Program Files directory.
```

binarymaster: I didn't tested the installer yet, but isn't it possible to just always extract to Program Files, so user won't be able to do a mistake here?

oleg-dubinskiy: Indeed. You're right. I forgot that it's already a custom installer, so we can change it as we need. Then this addition seems not required. :slightly_smiling_face:

HBelusca (approved):

PR #161, opened 2022-02-06
Added Adobe Flash Player 32

GatoGit21:
Added a Flash player that can run SWF files. It works on the nightly builds of ROS.


GatoGit21:
It contains a txt file with UTF-16 LE encoding and translated to 13 languages, and an icon.

oleg-dubinskiy:
> It contains a txt file with UTF-16 LE encoding and translated to 13 languages, and an icon.

I'm wondering why it's displayed as binary file on GitHub then...

GatoGit21:
> 

Yeah. That's weird, but here is the code if you don't trust it:


```
[Section]
Name = Adobe Flash Player 32
Version = 1.0.0
License = Freeware
Description = A 32 bits Flash player that can run SWF files.
Category = 12
URLSite = https://www.adobe.com/
URLDownload = https://fpdownload.macromedia.com/pub/flashplayer/updaters/32/flashplayer_32_sa.exe
SHA1 = 7736efc1c76e6a80132a22e1f9fc87a5884fa375
SizeBytes = 15986688
Icon = flashplayer.ico

[Section.0001]
Description = بت يمكنه تشغيل إثنان و ثلاثون بت يمكنها تنفيذ الملفات  SWF

[Section.0407]
Description = Ein 32-Bit-Flash-Player, der SWF-Dateien ausführen kann.

[Section.0a]
Description = Un reproductor de Flash de 32 bits que puede ejecutar archivos SWF.

[Section.040c]
Description = Un lecteur Flash 32 bits qui peut exécuter des fichiers SWF.

[Section.0410]
Description = Un lettore Flash a 32 bit in grado di eseguire file SWF.

[Section.0413]
Description = Een 32-bits Flash-speler die SWF-bestanden kan uitvoeren.

[Section.0415]
Description = 32-bitowy odtwarzacz Flash, który może uruchamiać pliki SWF.

[Section.0418]
Description = Un player Flash pe 32 de biți care poate rula fișiere SWF.

[Section.0419]
Description = 32-разрядный Flash-плеер, который может запускать SWF-файлы.

[Section.041f]
Description = SWF dosyalarını çalıştırabilen 32 bit Flash oynatıcı.

[Section.0422]
Description = 32-розрядний Flash-програвач, який може запускати файли SWF.

[Section.0425]
Description = 32-bitine Flash-mängija, mis suudab käitada SWF-faile.

[Section.0804]
Description = 可以运行 SWF 文件的 32 位 Flash 播放器。

[Section.0813]
Description = Een 32-bits Flash-speler die SWF-bestanden kan uitvoeren.
```

HBelusca:
> Yeah. That's weird, 
 
What you can do is to use an editor that can change the encoding to UTF-8 without BOM (for example, Notepad++) and update the file in this PR.

A separate suggestion: maybe add a screenshot of the app?

oleg-dubinskiy:
OK, thanks. I see that your file is indeed a correct txt file after downloading it. :slightly_smiling_face: 
But please specify which version of FP you've added: ActiveX, NPAPI or PPAPI? Just in case, we already have an existing entries for Flash Player. See fpnpapi.txt and fpppapi.txt appropriately. You may want to update (one of) them instead.

oleg-dubinskiy:
Ah, now I noticed that you added a standalone version (without browsers support). Then we may probably need to remove/update those old ones.

GatoGit21 (commented):
[flashplayer.txt](https://github.com/reactos/rapps-db/files/8011419/flashplayer.txt)
Added a screenshot of the app running on ROS. The image is located in Imgur.

GatoGit21:
> What you can do is to use an editor that can change the encoding to UTF-8 without BOM (for example, Notepad++) and update the file in this PR.

Does descriptions work with UTF-8 BOM in RAPPS? I can change it if it's needed.

> Ah, now I noticed that you added a standalone version (without browsers support). Then we may probably need to remove/update those old ones.

It's a player that runs without any browser. It's just a plain app. Should we take care of the other Flash downloads in other PR?

GatoGit21 (commented):
True. My bad.
I changed the file to UTF-8 encoding without BOM. It still includes the addition of a screenshot
[flashplayer.txt](https://github.com/reactos/rapps-db/files/8013949/flashplayer.txt)
.

learn-more:
You need to push your changes for us being able to see them, just updating files on your local PC is not going to update them here.

JoachimHenze:
We used to have two FlashPlayer setups in Rapps already in the past. NPAPI and PPAPI. Which one is that v32 that you propose here? Why don't you update one of the existing ones, but do create a 3rd one?

flashplayer.txt
```diff
@@ -0,0 +1,55 @@
+[Section]
+Name = Adobe Flash Player 32
+Version = 1.0.0
```
JoachimHenze, line 3: why v1.0.0 if it is v32 actually? Why not upgrading any of the two existing Flash setups in Rapps, but creating a 3rd one?

GatoGit21:
> You need to push your changes for us being able to see them, just updating files on your local PC is not going to update them here.

You're totally right. Now the txt is properly pushed into my fork. Sorry for any inconveniences I made.

> We used to have two FlashPlayer setups in Rapps already in the past. NPAPI and PPAPI. Which one is that v32 that you propose here? Why don't you update one of the existing ones, but do create a 3rd one?

The difference between those two and the app I'm proposing is that those are plugins ready to be installed on a browser, but Adobe Flash Player 32 is an independent player made by Adobe in the format of an exe file, that allows you to run your own downloaded SWF files without any need to run a browser or connecting to the Internet. AFP 32 also retains a series of settings such as zoom or quality. Needless to say that browsers (specially modern ones) doesn't support Flash Player plugins as easy as before, but AFP 32 gives the chance to the user to play those old games or media.

This combined with the fact that runs exactly the same as in Windows in ROS, I decided to publish this useful tool in RAPPS, and I publish it as another app because it's not a setup to install a plugin, like those two you're referring two.

flashplayer.txt
GatoGit21: Now it's fixed. Updated to 32.0.0.101 (2018).

JoachimHenze: Thx for fixing the version: Still the question: Why introducing a 3rd Flash player setup instead of updating the existing outdated ones?

PR #166, opened 2022-07-25
[RAPPS-DB] Update URL for UMing.

njlyf2011:
Accessing Gitee is faster than GitHub in China.
So I add the URL about UMing font.

### DIFF
--- a/arpl-uming.txt
+++ b/arpl-uming.txt
@@ -17,3 +17,5 @@ Description = AR PL UMing 字型是由臺灣文鼎科技開發股份有限公司
 
 [Section.0804]
 Description = AR PL UMing 字体是由台湾文鼎科技开发股份有限公司开发的一款开源字体。目前只支持中文环境。
+URLSite = https://gitee.com/njlyf2011/ttf-arphic-uming-reactos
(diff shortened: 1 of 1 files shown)


arpl-uming.txt
```diff
@@ -17,3 +17,7 @@ Description = AR PL UMing 字型是由臺灣文鼎科技開發股份有限公司
 
 [Section.0804]
 Description = AR PL UMing 字体是由台湾文鼎科技开发股份有限公司开发的一款开源字体。目前只支持中文环境。
+URLSite = https://gitee.com/njlyf2011/ttf-arphic-uming-reactos
+URLDownload = https://gitee.com/njlyf2011/ttf-arphic-uming-reactos/attach_files/1133634/download/uming-package-x86-32.exe
+SHA1 = 12acaa59129c9024d3a868b68d65b21e4c2fd3c9
```
HBelusca, line 22: If the SHA1 (and some other data) is the same as the common part in the general `[Section]`, I'm not sure these need to be re-specified there.

learn-more:
The script can't handle this case.
I'll try to fix this soon-ish.

learn-more:
Sorry for the delay, the script is fixed so I have merged this PR as well.

PR #173, opened 2022-10-06
Link update for four apps

simonelombardo:
1. **Hexedit1:** Main page for the version 1 has been removed from the host. Provided a web archive version of the URL Page.
2. **Photofiltre:** Provided alternative **not-**webarchive link for download since the executable version was removed (zip version still there). Same SHA-1.
3. **Scratch:** Cookie validation is required for download the software directly on MIT server. Provided a web archive version of download link. Same SHA-1.
4. **WMEncoder9:** File has been removed from the server. Provided an alternative **not-**webarchive link for the download. Same SHA-1.


JoachimHenze (approved):

photofiltre.txt
```diff
@@ -4,7 +4,7 @@ Version = 7.2.1
 License = Freeware
 Description = A complete image retouching program. This version comes with layers but is more limited than the Studio X version.
 Category = 3
-URLSite = http://www.photofiltre-studio.com/pf7-en.htm
+URLSite = http://202.129.240.234/Software/PhotoFiltre-7.2.1/pf7-setup-en-7.2.1.exe
```
binarymaster, line 7: This replaces a valid page URL with a download URL, while the actual download URL is not working.

PR #176, opened 2022-10-18
[CI] Improve validation python script

binarymaster:
- Add checks for section duplicates (happened in bcadc024)
- Add checks for required section keys (happened in 7001cf34)
- Add checks for value duplicates (happened in 7001cf34)
- Also fix URL to the RAPPS wiki

### DIFF
--- a/.github/validate-rapps.py
+++ b/.github/validate-rapps.py
@@ -29,6 +29,13 @@ ALL_KEYS = [
 ]
 
 
+REQUIRED_KEYS = [
+    b'Name',
+    b'Category',
+    b'URLDownload',
+]
+
+
 ALL_ARCH = [
     b'x86',
     b'amd64',
@@ -37,6 +44,10 @@ ALL_ARCH = [
     b'ia64',
     b'ppc',
 ]
+
+
+all_names = {}
+all_urls = {}
 
 
 HEXDIGITS = b'0123456789abcdef'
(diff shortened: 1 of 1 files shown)


binarymaster:
Yes, CI is expected to return failure now, because master is broken.
```
maelstorm.txt(1:0): Main section has no b'Category' key!
b'[Section]\r\n'
 ^
sopwith3.txt(2:0): Duplicate value found: b'Maelstorm, version 3.0.6'
b'Name = Maelstorm\r\n'
 ^
maelstorm.txt(2:0): First occurence:
b'Name = Maelstorm\r\n'
 ^
Please check https://reactos.org/wiki/RAPPS for details on the file format.
```

.github/validate-rapps.py
```diff
@@ -39,6 +46,12 @@
 ]
 
 
+all_name_lines = []
+all_url_lines = []
+all_names = []
+all_urls = []
```
learn-more, line 52: If you make these things a dictionary, you can lookup / use them easier.
f.e.:

```
# Define:
all_names = {}
# Lookup (note the 'None' default argument, without that this throws)
first = all_names.get(name, None)
# Use it
if first is not None:
    print(f'Duplicate shit, first occurrence: {first}')
else:
    all_names[name] = line
```

.github/validate-rapps.py
```diff
@@ -167,6 +185,14 @@ def __init__(self, fullname):
         self.filename = os.path.basename(fullname)
         self._sections = []
 
+    def verify_unique(self, reporter, lines, line, names, name):
```
learn-more, line 188: This does not need to be a class function, because all arguments are provided.

learn-more (commented):
Some minor comments

.github/validate-rapps.py
binarymaster: Addressed, thanks! :)

PR #180, opened 2022-11-13
[RAPPS-DB] Add Krita 3.2.1 to x86 and x64

julenuri:
Let's send this PR to sync with the ReactOS compatible Krita 3.2.1, last Windows XP compatible version. It's working nice. In the x64 version, have some font bug, that will be solved for sure in the near future. I will attach screenshots from other navigator if it allows me to do it, because Palemoon in ReactOS is not attaching the files in github.com xD

Edit: It doesn't need shims (in my experience) in Windows XP x64. However, if it does need any shim, lately could be applied (same that suffers the MDAC package). Screenshots of use without applying any shim (Luna theme :P).

Krita1
Krita2
Krita3

x64 screenshots about testing the software under ReactOS x64 (crt solved):

krita fixed 1
krita fixed 2

### DIFF
--- a/krita.txt
+++ b/krita.txt
@@ -2,7 +2,7 @@
 Name = Krita
 Version = 3.2.1
 License = GPL
-Description = Krita is an open-source graphic editor that allows to draw, paint and make 2D animation. Is based in KDE.
+Description = Krita is an open-source graphic editor that allows to draw, paint and make 2D animation. Is a part of KDE.
 Category = 3
 URLSite = https://krita.org
 URLDownload = https://download.kde.org/Attic/krita/3.2.1/krita-3.2.1-x86.zip
--- a/krita.txt
+++ b/krita.txt
@@ -1,0 +1,19 @@
+[Section]
+Name = Krita
+Version = 3.2.1
+License = GPL
+Description = Krita is an open-source graphic editor that allows to draw, paint and make 2D animation. Is based in KDE.
+Category = 3
+URLSite = https://krita.org
+URLDownload = https://download.kde.org/Attic/krita/3.2.1/krita-3.2.1-x86.zip
+SHA1 = 3f3ff0c2d971b8803fa978efd59f65aa7f8e8f1f
+SizeBytes = 71598080
+
+[Section.amd64]
+Version = 3.2.1
+URLDownload = https://download.kde.org/Attic/krita/3.2.1/krita-3.2.1-x64.zip
+SHA1 = cb85d630a8efdbc816296d757e05b1dc3d77943d
+SizeBytes = 100532224
+
+[Section.0a]
+Description = Krita es un programa de edición de imagen open source, que te permite dibujar, pintar y hacer animación 3D. Está basado en KDE.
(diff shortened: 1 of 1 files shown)


krita.txt
```diff
@@ -0,0 +1,19 @@
+[Section]
+Name = Krita
+Version = 3.2.1
+License = GPL
+Description = Krita is an open-source graphic editor that allows to draw, paint and make 2D animation. Is based in KDE.
```
oleg-dubinskiy, line 5:
```suggestion
Description = Krita is an open-source graphic editor that allows to draw, paint and make 2D animation. Is a part of KDE.
```

krita.txt
```diff
@@ -0,0 +1,19 @@
+[Section]
+Name = Krita
+Version = 3.2.1
```
oleg-dubinskiy, line 3:
```suggestion
Version = 3.2.1.0
```
To be more exact.

oleg-dubinskiy (commented):
I guess it requires enabling a shim compatibility mode before launching the installer/app? :thinking: We don't do it automatically via Rapps, so it still needs to be set manually. It will not launch correctly after downloading.
Btw last official Krita version supported by XP is 2.10.x, but yes, this one is also working fine (in compatibility mode). :slightly_smiling_face:

julenuri:
@oleg-dubinskiy the zipped one doesn't require any shim. However, the setup does. I included in this PR the zipped one.

oleg-dubinskiy:
> @oleg-dubinskiy the zipped one doesn't require any shim. However, the setup does. I included in this PR the zipped one.

Ah, okay then. Although the app itself still may need compatibility mode after extracting, I think it isn't a very big problem then.

julenuri:
@oleg-dubinskiy I meant, I haven't applied any shim in the version that is in the ticket. If we would wanted to attach the setup, then it might be needed. 
I can draw without big issues (I didn't try to make animations because I'm lazy). It does work in W2003 too, in both archs. Also, I'm not the only one that reports it working: https://www.reddit.com/r/krita/comments/6zq7n9/comment/dnfx5im/?utm_source=reddit&utm_medium=web2x&context=3

krita.txt
julenuri: In the software context, the webpage and the whole Krita context, that version is reported as 3.2.1, I will keep it like that. Searching in Google Krita 3.2.1.0 is not making any result.

krita.txt
```diff
@@ -0,0 +1,19 @@
...
+SizeBytes = 71598080
+
+[Section.amd64]
+Version = 3.2.1
+URLDownload = https://download.kde.org/Attic/krita/3.2.1/krita-3.2.1-x64.zip
+SHA1 = cb85d630a8efdbc816296d757e05b1dc3d77943d
+SizeBytes = 100532224
+
+[Section.0a]
+Description = Krita es un programa de edición de imagen open source, que te permite dibujar, pintar y hacer animación 3D. Está basado en KDE.
```
oleg-dubinskiy, line 19:
```suggestion
Description = Krita es un programa de edición de imagen open source, que te permite dibujar, pintar y hacer animación 3D. Es parte de KDE.
```

oleg-dubinskiy (commented):
One more suggestion and then LGTM from me. :slightly_smiling_face:

krita.txt
julenuri: Done.

oleg-dubinskiy (approved):

learn-more (approved):

PR #184, opened 2022-12-08
[RAPPS-DB] Update MS XML4 and XML6 with archive.org links

julenuri:


JoachimHenze (approved):
I double-checked: hashes are still fine.

PR #187, opened 2023-01-08
[RAPPS-DB] Deleted AC97 driver listing

hope2586:
I deleted the listing because ReactOS now has one built directly into itself (which can be improved if needed).


binarymaster:
0.4.14 release, which is still the latest one, doesn't have this driver.

hope2586:
I am closing this and will just wait to see if the driver gets added to the previous releases or not.

PR #192, opened 2023-02-19
[RAPPS-DB] Update K-Meleon's Listing

Daniel7689:


kmeleon.txt
```diff
@@ -1,35 +1,34 @@
...
+Version = 76.4.7
 License = GPL
-Description = An extremely fast, customizable and still lightweight web browser based on the Gecko layout engine.
+Description = An extremely fast, customizable and still lightweight web browser based on the Goanna layout engine.
 Category = 5
-URLSite = http://kmeleon.sourceforge.net
-URLDownload = http://download.sourceforge.net/project/kmeleon/k-meleon/75.1/K-Meleon75.1.exe
-SHA1 = a46ead69a6d18609b1f2e8ae60a3a5883e2e061e
-SizeBytes = 26570888
+URLSite = http://kmeleonbrowser.org/forum/read.php?19,154431
```
JoachimHenze, line 7: I checked: Using https here would not work. Not even from Windows, so this is fine.

JoachimHenze (approved):
Approved. Can you attach a picture of that version running on ros please?

JoachimHenze (changes requested):
Ah wait! You changed from a runnable .exe for the installer to a .7z. That will dramatically worsen the out-of-box-experience. Please change it back to something executable.

julenuri:
Here we have to take a choice for better out-of-box experience. We can stay in the last official version with Gecko, or jump to Goanna (best option since the old K-meleon needs the cypher update http://o.rthost.win/gpc/files1.rt/KM75.1-new-nss.7z for https and the goanna has it included). 
Both options are valid for me, but the bad thing is that roytam is not offering setups in the server, and only .7z, so we need to have an alternative installer.

PR #193, opened 2023-02-27
Let's add OnlyOffice

jeditobe:
OnlyOffice (formerly TeamLab), a powerful and free software office suite developed by Ascensio System SIA. Supports opening and saving documents in Microsoft Office formats.

It works in ReactOS quite nice and stable! Tested by me and some other people.

https://twitter.com/AgeNuclear/status/1628793519388205056

### DIFF
--- a/onlyoffice
+++ b/onlyoffice
@@ -1,6 +1,6 @@
 [Section]
 Name = OnlyOffice
-Version = 	7.3.0.184
+Version = 7.3.0.184
 License = AGPL-3.0-only
 Description = OnlyOffice (formerly TeamLab), a powerful and free software office suite developed by Ascensio System SIA. Supports opening and saving documents in Microsoft Office formats.
 Category = 6
--- a/onlyoffice
+++ b/onlyoffice
@@ -1,0 +1,13 @@
+[Section]
+Name = OnlyOffice
+Version = 	7.3.0.184
+License = AGPL-3.0-only
+Description = OnlyOffice (formerly TeamLab), a powerful and free software office suite developed by Ascensio System SIA. Supports opening and saving documents in Microsoft Office formats.
+Category = 6
+URLSite = https://www.onlyoffice.com/
+URLDownload = https://download.onlyoffice.com/install/desktop/editors/windows/distrib/onlyoffice/DesktopEditors_x86.exe
+SHA1 = 3a9b1cea5e088b3462dc0f942e34b3a2373be5ee
+SizeBytes = 206759344
+
+[Section.0419]
+Description = OnlyOffice (в прошлом Teamlab) — мощный офисный пакет с открытым исходным кодом, разработанный компанией Ascensio System SIA. Поддерживает открытие и сохрание документов в форматах Microsoft Office.


JoachimHenze (changes requested):
Please correct your previous PR first (the one with adding GPU-Z with the inaccessible URL before merging the next thing into rapps-DB)

learn-more:
> Please correct your previous PR first (the one with adding GPU-Z with the inaccessible URL before merging the next thing into rapps-DB)

That is not what reviews are for, a review is to be about this PR, not about a previous one.

jeditobe:
[JoachimHenze](https://github.com/JoachimHenze) I see no problems with #193, I think it should be approved

JoachimHenze (approved):
My concern has been addressed.
Sure this didn't invalidate the quality of PR193. But it is the kind of responsibility that we can/should expect from our contributors. As each app added into the DB will induce additional maintenance in the future. Therefore it makes sense to use a workflow that aims for completing step1 before step2 is approached. Thx for the fix. Approved now.

onlyoffice
```diff
@@ -0,0 +1,13 @@
+[Section]
```
julenuri, line 1: File doesn't have .txt extension, so it's not added to RAPPS

onlyoffice
```diff
@@ -0,0 +1,13 @@
...
+License = AGPL-3.0-only
+Description = OnlyOffice (formerly TeamLab), a powerful and free software office suite developed by Ascensio System SIA. Supports opening and saving documents in Microsoft Office formats.
+Category = 6
+URLSite = https://www.onlyoffice.com/
+URLDownload = https://download.onlyoffice.com/install/desktop/editors/windows/distrib/onlyoffice/DesktopEditors_x86.exe
+SHA1 = 3a9b1cea5e088b3462dc0f942e34b3a2373be5ee
+SizeBytes = 206759344
+
+[Section.0419]
+Description = OnlyOffice (в прошлом Teamlab) — мощный офисный пакет с открытым исходным кодом, разработанный компанией Ascensio System SIA. Поддерживает открытие и сохрание документов в форматах Microsoft Office.
```
julenuri, line 13: No newline at end of file

onlyoffice
JoachimHenze: I fixed that with https://github.com/reactos/rapps-db/commit/238ebb848f89d1852985d19f0f51f79b2298825e
Thx for the mentioning @julenuri

binarymaster:
I've fixed the download link to an XP compatible one in 671ea5cf405aea7101aed2f260e0e3eb3a9f964e.

There's also older version 6.1.0.94 [here](https://web.archive.org/web/20210214044140if_/http://download.onlyoffice.com/install/desktop/editors/windows/distrib/onlyoffice/DesktopEditors_x86_xp.exe) (207903672 bytes, SHA1 = `6e1e2aa707522bcbb98a589f426d00c4664f5944`), but I wasn't able to find version 7.3.0.184 which was initially added in this PR.

PR #195, opened 2023-03-03
GPU-Z Changed the link to a working one. 

jeditobe:
The name, size and hash sum of the file are the same as on the official website.


JoachimHenze (approved):
I do confirm, that the newly provided link works when entered into the browser on a Windows host.

PR #196, opened 2023-03-12
Add Clamwin without virus base

Artem13327:
Some data taken from clamwin.txt

The same Сlamwin only supplied without built-in virus databases


JoachimHenze:
What is the benefit of having that one without the signatures additionally?

learn-more:
Without something like that being indicated, people are going to get confused about which one to install,
so either one should explain why to install that one, or there should be one (and the other one be deleted).

JoachimHenze:
Closed due to inactivity

PR #206, opened 2023-10-24
Validation script enhancements

whindsaks:
- Python 3.4 compatible so that validate-rapps.py can be executed on ROS.
- Validate the license type field.
- Print a harmless warning if License is set to the useless value of "Unknown"

### DIFF
--- a/.github/validate-rapps.py
+++ b/.github/validate-rapps.py
@@ -45,6 +45,12 @@ ALL_ARCH = [
     b'ppc',
 ]
 
+LICENSE_TYPES = [
+    1, # Open source
+    2, # Freeware
+    3, # Trial/Demo
+]
+
 
 all_names = {}
 all_urls = {}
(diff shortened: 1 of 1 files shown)


JoachimHenze:
A pity that you bundled the license thing with that 2nd aspect of improving the compat to older Python.
I would really like to see that splitted into 2 different PRs! Aside from that, approved.

JoachimHenze (approved):

.github/validate-rapps.py
```diff
@@ -125,17 +131,19 @@ def _parse_section(self, reporter, stripped):
...
         if locale:
             if len(locale) not in (2, 4) or not all(c in HEXDIGITS for c in locale):
-                reporter.add(self, self._text.index(locale) + 1, f'Invalid locale{extra_locale}: "{locale}"')
+                reporter.add(self, self._text.index(locale) + 1,
+                             'Invalid locale{extra}: "{loc}"'.format(extra = extra_locale, loc = locale))
 
         if arch:
             if arch not in ALL_ARCH:
-                reporter.add(self, self._text.index(arch) + 1, f'Unknown architecture: "{arch}"')
+                reporter.add(self, self._text.index(arch) + 1, 'Unknown architecture: "%s"' % arch)
```
learn-more, line 146: Why are you using a different format method here?

whindsaks: Because the syntax is shorter and there is no ambiguity about which argument goes where.

learn-more (approved):

PR #219, opened 2024-01-09
[RAPPS-DB] Add Microsoft Visual C++ 2015-2019 Redistributable

TAN-Gaming:
Installs successfully on ReactOS without error.

VirtualBox_ReactOS_09_01_2024_17_28_43


learn-more:
Did you test the x64 version?
Because that setup looks like an x86 binary.

TAN-Gaming:
It seems the installer EXE of x64 version is 32-bit so I remove the section for now.

VirtualBox_ReactOS_x64_09_01_2024_19_51_12

learn-more (approved):

PR #222, opened 2024-02-03
[RAPPS-DB] Upgrade and add some emulators that are working under ReactOS

julenuri:
Upgrade some already had emulators:
PCSX ---> Upgraded to PCSX-R because the original link was down.
SNESx9 ---> Upgraded to the last version, minor improvements achieved. Add x64 option.

New software:
No$GBA, DeSmuME and Ideas ---> DS emulators. Some of them have really good behavior under VM, but under Real Hardware exposes some bugs we have.
Dolphin ---> Wii emu. Functional under 0.4.14 and previous, but we suffered some regression under 0.4.15 development. Fixed.
PCSX2-XP ---> PS2 emulator. Adapted version of the Vista+ version. When we will get the Vista+ functions, I will upgrade the version.
Project64 ---> N64 emulator. It's sometimes so random, and it reports that there is a bug at the start, but if you ignore the assert, you can play flawlessly under VM. Realhardware does weird things.

After this PR I will try to add some PR with other GPL emulator of Sega, PS3, WiiU and 3DS machines.

I could add some screenshots if you want.


desmume.txt
```diff
@@ -0,0 +1,20 @@
+[Section]
+Name = DeSmuME
+Version = 0.9.11
+License = GNU GPL v2
+Description = DeSmuME is one of the most popular Nintendo DS emulator. Needs Microsoft Visual C++ 2010 Redistributable.
+Category = 4
+URLSite = http://desmume.org/
```
JoachimHenze, line 7:
```suggestion
URLSite = https://desmume.org
```

pcsx2-xp.txt
```diff
@@ -0,0 +1,15 @@
+[Section]
+Name = PCSX2 (XP Compatible)
+Version = 1.6.0
+License = GNU GPL v2
+Description = PCSX2 is a Playstation 2 emulator. It needs Microsoft Visual C++ 2013 redistributable.
+Category = 4
+URLSite = http://neonfloppy.sytes.net/projects/pcsx2-xp/
+URLDownload = http://neonfloppy.sytes.net/projects/pcsx2-xp/releases/1.6.0/pcsx2-xp-binaries.7z
```
JoachimHenze, line 8: I double-checked already: http**s** doesn't work for those 2 links on Windows.

julenuri: So I suggest two options:
1. Trust the http anyway until we have support for Vista+ functions and then update the links to the common normalized version.
2. Change the links through Webarchive, that drops the speed but is more trusty.

JoachimHenze: No change requested necessarily. I just wrote that, to prevent others recommending to change the URL to http**s** (without them testing the result).
Our additional hash should guarantee integrity also when transferring via http.

JoachimHenze (approved):

PR #227, opened 2024-03-08
[RAPPS-DB] Add Fileoptimizer

marjuanm:
FileOptimizer is an application to reduce unnecessary weight of files, I have only tested it on image files and it works very well in Reactos.


marjuanm:
I'm adding some screenshots of FileOptimizer running in Reactos.
Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4

marjuanm:
Before and after compressing the graphic files, as we see, the weight of the file has been reduced a little without damaging the image or causing it to lose its quality.
before
after

JoachimHenze (approved):
Approved. I will merge it in a week if nobody will merge it earlier than that.

learn-more (approved):

marjuanm:
thanks @JoachimHenze JoachimHenze

PR #232, opened 2024-03-13
[RAPPS-DB] Add ReactOS Wallpaper Exposition v5 (roswallpexpo)

katahiromz:
Retrial of #230. This is a wallpaper set by JIRA user `fastcom`.

https://jira.reactos.org/browse/CORE-12639

apple

leaves

plain

sky

sunset

VirtualBox_ReactOS_17_03_2024_13_45_26

### DIFF
--- a/roswallpexpo.txt
+++ b/roswallpexpo.txt
@@ -1,10 +1,16 @@
 [Section]
 Name = ReactOS Wallpaper Exposition
-Version = 5
+Version = 6
 License = CC BY
-Description = A wallpaper set (apple, leaves, sky etc.), by fastcom.
+Description = A wallpaper collection (apple, leaves, sky etc.), by fastcom.
 Category = 16
-URLSite = https://jira.reactos.org/browse/CORE-12639
-URLDownload = https://jira.reactos.org/secure/attachment/67625/RosWallpExpo-v5.exe
-SHA1 = 2b74fb974778aa51a8695464cb695bfd0252643f
-SizeBytes = 2211512
+URLSite = https://github.com/katahiromz/RosWallpExpo
+URLDownload = https://github.com/katahiromz/RosWallpExpo/releases/download/v6/RosWallpExpo-v6.zip
+Installer = Generate
+SHA1 = 7a493c8687c462ec9eb75daacfd6ca9e1233b08e
+SizeBytes = 1699194
+
+[Generate]
+Files=RosWallpExpo-v6\*
+Dir=%SystemRoot%\Web\Wallpaper
+Lnk=!
--- a/roswallpexpo.txt
+++ b/roswallpexpo.txt
@@ -1,0 +1,10 @@
+[Section]
+Name = ReactOS Wallpaper Exposition
+Version = 1
+License = CC BY
+Description = A wallpaper set (apple, leaves, sky etc.), by fastcom.
+Category = 16
+URLSite = https://jira.reactos.org/browse/CORE-12639
+URLDownload = https://jira.reactos.org/secure/attachment/67588/ReactOSWallpaperExposition-v1.zip
+SHA1 = 3a997c41b26e90bcdc0adaf46c60c89f91ef7b5d
+SizeBytes = 999104
--- a/roswallpexpo.txt
+++ b/roswallpexpo.txt
@@ -1,10 +1,10 @@
 [Section]
 Name = ReactOS Wallpaper Exposition
-Version = 1
+Version = 5
 License = CC BY
 Description = A wallpaper set (apple, leaves, sky etc.), by fastcom.
 Category = 16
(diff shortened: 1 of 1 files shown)


roswallpexpo.txt
```diff
@@ -0,0 +1,10 @@
+[Section]
+Name = ReactOS Wallpaper Exposition
+Version = 1
+License = CC BY
+Description = A wallpaper set (apple, leaves, sky etc.), by fastcom.
+Category = 16
+URLSite = https://jira.reactos.org/browse/CORE-12639
+URLDownload = https://jira.reactos.org/secure/attachment/67588/ReactOSWallpaperExposition-v1.zip
```
learn-more, line 8: Does this work in ReactOS ?

katahiromz: image
Oh, unsupported compression method. I will retry.

katahiromz: And we didn't support JPEG wallpapers.

katahiromz: VirtualBox_ReactOS_17_03_2024_13_45_26
Now it works!

HBelusca:
@katahiromz I would suggest we talk with @ColinFinck so that we can have an _**official**_ repository for wallpapers, themes and stuff under our github.com/reactos .

marjuanm:
Hello:

  One question: is this intended to have "collections" of wallpapers for Reactos? I think that putting those wallpapers in zips and having the user download them doesn't tell the user much, a zip in the repository should containing a preview view with images inside. Or Reactos should had a utility that allowed access to a list of wallpapers with their respective preview and then the user would download the one they wanted to put them as the background on their desktop.

katahiromz:
> is this intended to have "collections" of wallpapers for Reactos?

Yes, it is.

> I think that putting those wallpapers in zips and having the user download them doesn't tell the user much, a zip in the repository should containing a preview view with images inside.

Now, this package is not a ZIP. But an installer of an EXE file.

> Or Reactos should had a utility that allowed access to a list of wallpapers with their respective preview and then the user would download the one they wanted to put them as the background on their desktop.

Rapps didn't support screenshot preview.

marjuanm:
This proposal sounds interesting and seeing it carried out I think it would give Reactos a good boost. I remember that one of the attractions of Windows Vista was its collection of wallpapers (beautiful landscapes, views of space, etc.) and of course Aero theme.

However, I was referring to developing an app that would download those wallpapers from a folder on the network or a repository and show a preview of them, if that is not possible or is not in the plans for Reactos and you are going to continue using the common repository for all applications, I think that from time to time there will be a need to update that app because it is likely that the number of wallpapers will be renewed, right?

JoachimHenze:
@katahiromz would it be possible to pack the wallpapers from rosapps module into such a downloader as well?
I do kind of dislike that they do bloat the release-isos. Rapps is typically just one click away.

roswallpexpo.txt
```diff
@@ -0,0 +1,10 @@
+[Section]
+Name = ReactOS Wallpaper Exposition
+Version = 5
+License = CC BY
+Description = A wallpaper set (apple, leaves, sky etc.), by fastcom.
+Category = 16
+URLSite = https://jira.reactos.org/browse/CORE-12639
+URLDownload = https://jira.reactos.org/secure/attachment/67625/RosWallpExpo-v5.exe
```
JoachimHenze, line 8: I am not sure whether it would be wise to apply additional traffic onto our JIRA servers by adding such links to JIRA into rapps. I think the file should be placed on github for example. Which does offer servers with high availability, and stable links. And we don't have to pay/maintain that infrastructure ourself.

katahiromz:
> @katahiromz would it be possible to pack the wallpapers from rosapps module into such a downloader as well?
I do kind of dislike that they do bloat the release-isos. Rapps is typically just one click away.

I can make `modules\wallpapers` as a Rapps application. It reduces 7MB download size.

roswallpexpo.txt
katahiromz: We can use my private GitHub Release page or our official Release page for downloadinp. Official page is the better but I have no rights to modify our official Release page.

JoachimHenze: Our release-page https://github.com/reactos/reactos/releases shouldn't get messed with "unrelated files" sprinkled in between our official ros-release-announcements.
Optimal for such stuff would be to create some project like https://github.com/reactos/hosted-files
As long as something like that doesn't exist (I do understand you don't have the rights for such a thing),
I do think **your private GitHub Release page is a sane alternative, yes**.
We do already have a lot of rapps links of the form https://github.com/someContributor/SomeProject

katahiromz:
Updated.

katahiromz:
Retired.

PR #211, opened 2023-11-14
[RAPPS-DB] Fixing icons and adding some software that is x64 compatible - Update november 1

julenuri:
In this PR:

I fix some broken links (abiword website is down and I have pending the PCSX).
Fix the ff-mpeg commit (I am not sure why, but it's broken on Git).
Add Musescore, Notepad++ and SuperTux x64 binaries.
Update VLC to the last compatible version.
Update Abyss Web server X1 to the last compatible version.
Add Blender in both installable msi.

VirtualBox_ReactOS x64_14_11_2023_17_39_48
VirtualBox_ReactOS master_14_11_2023_14_44_57
VirtualBox_ReactOS x64_14_11_2023_22_36_17
VirtualBox_ReactOS x64_14_11_2023_22_14_32
VirtualBox_ReactOS x64_14_11_2023_18_04_42
VirtualBox_ReactOS x64_14_11_2023_17_54_05

PR #233, opened 2024-03-21
[RAPPS-DB] Fix PNG icons for renpy.ico, sublime_text.ico, and winrar.ico

Doug-Lyons:
This is a fix of [PR#211](https://github.com/reactos/rapps-db/pull/211) and commit [d2a7485](https://github.com/reactos/rapps-db/commit/d2a7485)

Three icons were included with this commit which were renpy.ico, sublime_text.ico and winrar.ico.
These were not actual ICO files, but actually PNG files that had had their extensions changed to ICO.

This commit replaces these bad icon files with the correct ICO files with their appropriate icon headers.
These will not display correctly in ReactOS, but there is work in progress to display these soon.


learn-more:
Thanks @Doug-Lyons, after this PR is merged we can also merge the updated validator, so that this does not happen again:
#234

learn-more (approved):

Doug-Lyons:
> > > These will not display correctly in ReactOS, but there is work in progress to display these soon.
> 
> So they do still have png data inside?

Yes. They will make good test examples to verify our PNG changes work after they are merged later.
In the meantime, they should trigger my debug messages telling us that they have PNG's embedded.
They are not helpful as they are now. They are not displayable by any OS or ReactOS.

JoachimHenze (changes requested):
ChangesRequested: I don't think this PR makes sense.
To underline my concerns: Here is a pseudo-ico-file that I created with Win8.1 paint, I saved a 16x16bmp as png. [test.zip](https://github.com/reactos/rapps-db/files/14717058/test.zip)
And then I renamed that png to .ico.
My Win8.1 does properly display its "icon" on my desktop.
I deduce from that, that at least a modern Windows should be able to use a file as pseudo-ico after just renaming a png file.

And if I try the same on my WinXP, it also works,(ofc not with the renamed png), but it works if I do use a bmp and rename that to ico.

To sum it up:
Windows (both NT5 and NT6) seem to parse an ico more cleverly than ros, and finds also other image formats inside, even without an extra-header.

Here are 2 ways forward that would make sense imho:
1.) the conservative approach would replace those pngs with proper "bmp-icos". That is what our target 2k3sp2 would understand as well.
2.) The more aggressive approach to the future: The files are left like they are today, and ros is pimped with more NT6-stuff that would allow to display the renamed files fine.

Instead adding that header in this PR would neither satisfy our target (avoiding NT6), nor would it be really enforcing us to develop something that is really WindowsNT6 compatible. And as an extra-malus: the pseudo-icos would increase in size.

binarymaster (approved):

binarymaster:
Even if Windows is able to use bitmaps as icons, it shouldn't move us away from keeping quality standards.

JoachimHenze:
Quality standards would mean that we try to achieve what Windows can achieve.
This PR does neither fix the icon displaying, nor does it help ReactOS evolve. The only thing it does is wasting our time on the noise in the rapps DB. By making it even larger. This PR does not have any immediate gain. It just makes it easier and even more tempting for us to accumulate even more workarounds. Workarounds are no quality standard.

PR #234, opened 2024-03-21
[CI] Validate uploaded icons

learn-more:


binarymaster (approved):

JoachimHenze (approved):

whindsaks:
.ico files just start with 3 zero bytes or something like that, would be trivial to do without external tools if all one cares about is not being confused with a PNG or jpg.

whindsaks:
> Does this updated script no longer work for you?

No, not directly no.

```
python -m pip install --upgrade pip
python -m pip install python-magic
python -m pip install python-magic-bin
.github\validate-rapps.py
```
and it then fails with a missing VC runtime error.

All you need is
```
import struct
f = open("winfile.ico", "rb")
if struct.unpack("<H", f.read(2))[0] == 0:
	if struct.unpack("<H", f.read(2))[0] == 1:
		if struct.unpack("<H", f.read(2))[0] > 0:
			print("Smells like a .ico")
f.close()
```

learn-more:
@whindsaks Thanks for the suggestion, I have (finally) updated the script to manually parse the .ico without dependency.

PR #235, opened 2024-04-30
[RAPPS-DB] Add Firefox 52.9 ESR

TAN-Gaming:
As of commit [24a56f89](https://github.com/reactos/reactos/commit/24a56f89ab5cd529a724f1b5de9ce239e7170b6a) , Firefox 52.9.0 ESR works on ReactOS without requiring `IgnoreManifestCompatVersion` compatibility settings.

### DIFF
--- a/firefox-esr.txt
+++ b/firefox-esr.txt
@@ -1,0 +1,11 @@
+[Section]
+Name = Mozilla Firefox 52 ESR
+Version = 52.9.0
+License = MPL/GPL/LGPL
+Description = A popular Web browser. Free and open source, made by a global community.
+Category = 5
+URLSite = https://www.mozilla.org/en-US/
+URLDownload = https://ftp.mozilla.org/pub/firefox/releases/52.9.0esr/win32/en-US/Firefox%20Setup%2052.9.0esr.exe
+SHA1 = 705240c720a631c5b7c70ef41f5af0f272721ea9
+SizeBytes = 45470192
+Icon = firefox.ico


JoachimHenze (approved):
Approved. We might consider stripping the FF48.0.2 within forefox.txt from the rappsDB now, as it doesn't provide anything superior over 52.9 ESR, as far as I am aware of. And FF48 tries to download and update itself to FF52 anyway upon first start.

We should keep the "3.6" / "28 ESR" and "52.9 ESR" though. Each of those is kind of valuable.

firefox-esr.txt
```diff
@@ -0,0 +1,11 @@
+[Section]
+Name = Mozilla Firefox 52 ESR
+Version = 52.9.0
+License = MPL/GPL/LGPL
+Description = A popular Web browser. Free and open source, made by a global community.
```
JoachimHenze, line 5: It may make sense to add a sentence `Scores xxx/555 on https://html5test.opensuse.org` where xxx would have to be tested though.

TAN-Gaming:
> We might consider stripping the FF48.0.2 within forefox.txt from the rappsDB now, as it doesn't provide anything superior over 52.9 ESR

I think we should update it to the latest compatible version and keep firefox.txt as is (for non-ESR version).

> And FF48 tries to download and update itself to FF52 anyway upon first start.

You can disable that by not installing the maintenance service.

firefox-esr.txt
TAN-Gaming: Maybe in the later PR, once and for all.

PR #237, opened 2024-05-06
[RAPPS-DB] Update icon for all Firefox versions

TAN-Gaming:
Capture_RAPPS


oleg-dubinskiy (commented):
Just my two suggestions:
1. Maybe we can use a newer Firefox icon for 48 and 52 ESR versions, those have it originally? You can simply extract and use it from the firefox.exe binary via any resource editor, since it's GPL. :wink: 
2. Since we already have a newer 52 ESR version working in master, we might want to remove an older 48 version at all (outside of this PR), since this was a minor maintenance release which did not introduce any major changes comparing to some another versions. :slightly_smiling_face:

JoachimHenze (approved):

TAN-Gaming:
> Since we already have a newer 52 ESR version working in master, we might want to remove an older 48 version at all (outside of this PR), since this was a minor maintenance release which did not introduce any major changes comparing to some another versions.

I think we should update it to the latest compatible non-ESR version and keep firefox.txt in the database as is (for non-ESR version) so the user will be able to choose between ESR/non-ESR version.

TAN-Gaming:
Updated:
CaptureRAPPS2

Let's go.

PR #246, opened 2024-05-29
[RAPPS-DB] Add Dependency Walker

whindsaks:


dependencywalker.txt
```diff
@@ -0,0 +1,16 @@
...
+LicenseType = 2
+Category = 7
+URLSite = https://dependencywalker.com
+Version = 2.2
+URLDownload = https://dependencywalker.com/depends22_x86.zip
+SizeBytes = 468618
+Installer = Generate
+
+[Section.amd64]
+URLDownload = https://dependencywalker.com/depends22_x64.zip
```
HBelusca, line 13: No SizeBytes for the x64 entry?

HBelusca (commented):
Can you remind us what category 7 is? (i.e. where the tool will land into)

whindsaks:
> Can you remind us what category 7 is?

Development

dependencywalker.txt
HBelusca: Also we usually add a SHA entry to verify the download has been done correctly. This seems missing.

dependencywalker.txt
```diff
@@ -0,0 +1,16 @@
+[Section]
+Name = Dependency Walker
+Description = A free utility that scans any 32-bit or 64-bit Windows module (exe, dll, ocx, sys, etc.) and builds a hierarchical tree diagram of all dependent modules.
```
JoachimHenze, line 3:
```suggestion
Description = MsDepends is a free utility that scans any 32-bit or 64-bit Windows module (exe, dll, ocx, sys, etc.) and builds a hierarchical tree diagram of all dependent modules.
```
I want to add the term "MSDepends" here, because people might enter that into the rapps searchbox! So it will be found then also.

dependencywalker.txt
```diff
@@ -1,16 +1,19 @@
...
 URLDownload = https://dependencywalker.com/depends22_x86.zip
-SizeBytes = 468618
+SHA1 = bfec714057e8449b0246051be99ba46a7760bab9
+SizeBytes = 610769
 Installer = Generate
 
 [Section.amd64]
 URLDownload = https://dependencywalker.com/depends22_x64.zip
+SHA1 = 4831d2a8376d64110ff9cd18799fe6c69509d3ea
+SizeBytes = 468618
```
JoachimHenze, line 16: A bit surprising that the x64 has smaller filesize than the x86. That is pretty uncommon. No biggy, I am just expressing my mind here in case you swapped the numbers accidentally. If not, then fine.

whindsaks: Yes, it is correct. The 64-bit depends.exe is a lot smaller, perhaps missing Windows 9x code? depends.dll is larger.

PR #254, opened 2024-07-10
Add Safari App

Usinganame:


Usinganame:
Closed as github has to worst UI known to man and I don't have a clear idea of what different buttons do or they do two different things

binarymaster:
> github has to worst UI known to man

That's debatable.

PR #255, opened 2024-07-10
Add Safari 5.1.7

Usinganame:
Safari 5.1.7 is the latest version of Safari to support Windows


safari.txt
```diff
@@ -0,0 +1,10 @@
+[Section]
+Name = Safari
+Version = 5.1.7
+License = EULA
```
binarymaster, line 4:
```suggestion
License = EULA
LicenseType = 2
```
https://reactos.org/wiki/RAPPS#LicenseType

safari.txt
```diff
@@ -0,0 +1,10 @@
...
+Description = Apple Safari for Windows version 5.1.7
```
binarymaster, line 5:
```suggestion
Description = Apple Safari browser for Windows.
```
As the version is already in the other field.

safari.txt
```diff
@@ -0,0 +1,10 @@
...
+Category = 5
+URLSite = https://archive.org/details/safari-5.1.7-windows
+URLDownload = https://archive.org/download/safari-5.1.7-windows/5.1.7_SafariSetup.exe
+SizeBytes = 38494576
+Icon = safari.ico
```
binarymaster, line 10: You can omit this field, as the icon filename matches the txt filename.

https://reactos.org/wiki/RAPPS#Icon

```suggestion
```

safari.txt
```diff
@@ -0,0 +1,10 @@
+[Section]
+Name = Safari
+Version = 5.1.7
+License = EULA
+Description = Apple Safari for Windows version 5.1.7
+Category = 5
+URLSite = https://archive.org/details/safari-5.1.7-windows
+URLDownload = https://archive.org/download/safari-5.1.7-windows/5.1.7_SafariSetup.exe
+SizeBytes = 38494576
```
binarymaster, line 9: Would be nice to add SHA1 checksum here.

https://reactos.org/wiki/RAPPS#SHA1

binarymaster (commented):
Have you tested it in ReactOS? For me it installs, but runs unbelievably slow (barely usable).

VirtualBox_ReactOS_11_07_2024_13_37_58

Usinganame:
I have tested it, but it was under VirtualBox and as i can't install the necessary drivers it runs slow either way and i don't have a spare computer laying around. I just assumed that it was slow either way, like Chromium or Mypal

JoachimHenze (changes requested):
It runs unbearably slow due to existing bugs in ReactOS. That has nothing to do with the VBox emulation.
Since our common practice is to only add applications to rapps that are known to work in an acceptable manner, I do suggest not merging this PR until somebody would be able to name at least one website, which can only be opened by this browser, but not a single other available browser option for ros.
Either that, or we should wait with the merging until the slowness-bug in ros will have been fixed already.
I don't see much use in adding a lot of applications into rapps, that don't work well yet.

The last bug that we (think we) fixed in Safari-context was https://jira.reactos.org/browse/CORE-13125 but that doesn't deal specifically with the slowness. If there is not a slowness-bug filed in JIRA already that I missed, then that should be created **now** and be linked here in this conversation.

PR #271, opened 2024-09-05
[RAPPS-DB] Add missing cheat engine icon and turkish description

scropool:
I think it turned out pretty good :D


julenuri:
imagen

It requires few bytes and has a decent look, shipping it.

PR #272, opened 2024-09-06
[RAPPS-DB] Add HWMonitor 1.52.0

scropool:
image

unfortunately it only works on 32 bit and this version :(


julenuri (approved):
For me this is a clear yes to rapps-db.

learn-more (approved):

PR #277, opened 2024-10-11
Add NVIDIA Geforce 181.22 (WinXP driver)

amir16yp:
Added nvidia forceware 181.22. was tested on many supported cards and had successful results with 2D acceleration

### DIFF
--- a/nvidiaforceware.txt
+++ b/nvidiaforceware.txt
@@ -1,0 +1,9 @@
+[Section]
+Name = NVIDIA Forceware driver
+Version = 181.22
+License = Unknown
+Description = Driver for GeForce 6, 7, 8, 9, and 200-series NVIDIA GPUs
+Category = 13
+URLDownload = https://us.download.nvidia.com/Windows/181.22/181.22_geforce_winxp_32bit_english_whql.exe
+SizeBytes = 74060000
+SHA1 = ffea1a2c9f8e13c0393259eda686cc4f13b72cc3
--- a/nvidiaforceware.txt
+++ b/nvidiaforceware.txt
@@ -2,7 +2,7 @@
 Name = NVIDIA Forceware driver
 Version = 181.22
 License = Unknown
-Description = Driver for GeForce 6, 7, 8, 9, and 200-series NVIDIA GPUs
+Description = Driver for GeForce 6, 7, 8, 9, and 200-series NVIDIA GPUs.
 Category = 13
 URLDownload = https://us.download.nvidia.com/Windows/181.22/181.22_geforce_winxp_32bit_english_whql.exe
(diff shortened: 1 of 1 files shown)


HBelusca:
Do you have a screenshot that shows that it indeed works in ReactOS?

amir16yp:
> Do you have a screenshot that shows that it indeed works in ReactOS?

yes. here's an ROS rig that i tested, used this exact driver: https://x.com/reactos/status/1704856015345455578?s=20

additionally there are many entries on the ROS wiki regarding GPUs that were tested with this driver: https://reactos.org/wiki/Supported_Hardware/Video_cards

nvidiaforceware.txt
```diff
@@ -0,0 +1,9 @@
+[Section]
+Name = NVIDIA Forceware driver
+Version = 181.22
+License = Unknown
+Description = Driver for GeForce 6, 7, 8, 9, and 200-series NVIDIA GPUs
```
HBelusca, line 5:
```suggestion
Description = Driver for GeForce 6, 7, 8, 9, and 200-series NVIDIA GPUs.
```

nvidiaforceware.txt
```diff
@@ -0,0 +1,9 @@
+[Section]
```
HBelusca, line 1: Please ensure that your file is encoded in UTF-8 without BOM, with CR-LF endings.

nvidiaforceware.txt
```diff
@@ -0,0 +1,9 @@
+[Section]
+Name = NVIDIA Forceware driver
+Version = 181.22
+License = Unknown
```
HBelusca, line 4: I guess that this would be the license
https://www.nvidia.com/en-us/drivers/geforce-license/

nvidiaforceware.txt
binarymaster: CI is passing.

JoachimHenze (approved):
Approved.

julenuri:
I will add the 64 bits version too, only if anybody is brave enough to test it under Real Hardware. This PR is be a pass for me. Also adding a ico will be ok.

amir16yp:
> I will add the 64 bits version too, only if anybody is brave enough to test it under Real Hardware. This PR is be a pass for me. Also adding a ico will be ok.

there's an already an ico for the other (older) nvidia display driver. would it be a good idea to use it?

julenuri:
> > I will add the 64 bits version too, only if anybody is brave enough to test it under Real Hardware. This PR is be a pass for me. Also adding a ico will be ok.
> 
> there's an already an ico for the other (older) nvidia display driver. would it be a good idea to use it?

I will add a "more recent one". Don't worry. :)

PR #284, opened 2024-11-15
[RAPPS-DB] Add some icons - Part III

crusedeelo:
only some of the icons i will add more

here is the list i added last:

```
mozilla sunbird > OK 1 +
tightvnc > OK 2 +
celeste > OK 3 +
ant renamer > OK 4 +
autoruns > OK 5 +
AlReader 2 > OK 6 +
clickntype > OK 7 +
cURL > OK 8 +
directx > OK 9 + Added 2 icons for this feb 2010 and latest
dplus browser > OK 10 +
disk internals linux reader > OK 11 + 
double commander > OK 12 +
effective file search > OK 13 +
```

dvd write now > it caused to crash reactos i think the reason his own driver xD::D i cant reach icon

to be continued another day..


julenuri:
Two bad icon headers...

PR #286, opened 2024-12-01
[RAPPS-DB] Add TDM-GCC 10.3.0 (32-bit)

TAN-Gaming:
64-bit version doesn't work right now because the setup EXE is 32-bit.

GCC version test:
VirtualBox_ReactOS_01_12_2024_23_41_43

Test building a simple Win32 program:
VirtualBox_ReactOS_01_12_2024_23_44_03


HBelusca (approved):

TAN-Gaming:
Let's go.

PR #288, opened 2025-01-11
[RAPPS-DB] Remove Spotify 1.0.29.92

oleg-dubinskiy:
Remove Spotify 1.0.29.92 from Rapps database because this offical client for Windows XP/Server 2003 has been dropped and is no longer supported by Spotify starting from December 2022. So now it has exactly zero functionality for us and hence is useless (except for regression testing probably). It does not allow to login with credentials and can't be used anymore.
I will replace it by SpotifyXP, which is an unofficial (3rd party) client written in Java, and it works in ReactOS, as it's made especially for XP/2003 and is the only way to listen Spotify music on these OSes. After recent bugfix it works in ReactOS too.
Here's an error which appears when trying to login from this old Spotify client. Earlier it had another error "This version of Spotify is restricted to Premium users only", but now, after Spotify has dropped login via password and introduced a new authentification via web browser, it just says "Wrong e-mail and/or password". However I tried many times: my entered credentials are correct because they allow me to login properly via web-browser on Spotify website (e. g., when logging via OAuth method from SpotifyXP). But, as you can see, they don't work with an old client (anymore):
Spotify-1 0 29 92


julenuri:
Yeah, I do agree. And actually, it's nothing else than a Chromium client with little flavor so...

whindsaks (approved):

PR #291, opened 2025-01-21
[RAPPS-DB] Convert .zip to generated installers

whindsaks:
Automatically generate installers for .zip files.

Notes:
 - Folder2Iso does not have a valid 32-bit download, this does not fix that.
 - There are still several remaining packages that still are plain .zip files for various reasons.


Folder2Iso.txt
```diff
@@ -9,6 +9,13 @@ URLDownload = https://ia802807.us.archive.org/27/items/folder2iso_202001/Folder2
 SHA1 = 07c1641ada57f31964eb4327d5753ab310fc5975
 SizeBytes = 2736100
 Icon = folder2iso.ico
+Installer = Generate
+
+[Section.amd64]
+URLDownload = https://ia802807.us.archive.org/27/items/folder2iso_202001/Folder2Iso.zip
```
HBelusca, line 15: Same URL for both "default" and x64? If so then the Section.amd64 is redundant.

Cygne.txt
```diff
@@ -9,6 +9,11 @@ URLDownload = http://cygne.emuunlim.com/files/cygne21a-win.zip
 SHA1 = 4119a3b1f8cba6c663e04d6f4eb60116485a112a
 SizeBytes = 100197
 Icon = cygne.ico
+Installer = Generate
+
+[Generate]
+DelDirEmpty = SRAM
```
HBelusca, line 15: Just to be sure, this list of directories are temporary ones created by the software?

precisecalc.txt
```diff
@@ -6,8 +6,13 @@ Description = A free and open source scientific calculator for Windows.
...
 URLDownload = http://download.sourceforge.net/preccalc/preccalc-32bit.zip
-SHA1 = 6a2711d3c72708f3876ac128fcd1f19a5145d77a
-SizeBytes = 192127
+SHA1 = 0c2c48b3ffc3a6f0faef74c539d3bf9f2cbedcc4
+SizeBytes = 208036
+Installer = Generate
+
+[Generate]
+DelReg = HKCU\Software\Petr Lastovicka\calc
+DelRegEmpty = HKCU\Software\Petr Lastovicka
```
HBelusca, line 15: Side-question: does the parser allows for specifying values within quotes (like: `DelRegEmpty = "HKCU\Software\Petr Lastovicka"`) yet doing the correct thing?

restoration.txt
```diff
@@ -3,10 +3,11 @@ Name = Restoration
 Version = 3.2.13
 Description = Restoration is a freeware, portable app that can restore files accidentally deleted from the recycle bin.
 Category = 12
-URLDownload = https://files1.majorgeeks.com/10afebdbffcd4742c81a3cb0f6ce4092156b4375/drives/Restoration.zip
+URLDownload = https://web.archive.org/web/2024if_/https://files1.majorgeeks.com/acd9affbbee5e4c28a3a457dcf138bb1259439e4/drives/Restoration.zip
```
HBelusca, line 6: Different "hash" in the URL, but the resulting file's SHA1 and size (below) are the same?

xbgpack1.txt
```diff
@@ -8,6 +8,12 @@ URLSite = https://web.archive.org/web/20190820014045if_/http://xcallono.cf/
 URLDownload = https://web.archive.org/web/20221016174911if_/https://chat.reactos.org/files/1rxc6869z7rfdqdoiyq1wdqjoe/public?h=nWi06_4V2YHASeWtzUt_rWQMKKcOESXXMud7xwwHJVM&f=/Mario%2064%20Backgrounds.zip
```
HBelusca, line 8: huhu I think we should archive this file elsewhere as well.

HBelusca (commented):
I skimmed through your proposed changes, I haven't found any weird things per se (except the apparent redundant `Section.amd64` for one entry).
I guess that if one wants to accurately double-check your proposal, one would have to systematically install each individual zip-provided app to verify that the generated installer will correctly delete all the created (incl. extra temporary) files.

Folder2Iso.txt
whindsaks: See the my note. Folder2Iso incorrectly links to the 64-bit version and I could not find a 32-bit version. Section.amd64 is now correct. Hopefully non-onedrive link can be found for the 32-bit version at some point.

Cygne.txt
whindsaks: Correct.

Folder2Iso.txt
HBelusca: ah, i understand.

precisecalc.txt
whindsaks: No, `|` is the separator, it does not care about spaces. Quotes are problematic because some .ini parsers will remove them (not sure if Rapps does).

xbgpack1.txt
whindsaks: Well, it's already on archive.org, what more do you want?

whindsaks:
> I guess that if one wants to accurately double-check your proposal, one would have to systematically install each individual zip-provided app to verify that the generated installer will correctly delete all the created (incl. extra temporary) files.

I pretty much tried every installer in this PR. I started the program and then tried to configure its settings dialog if it has one. I did not use the programs a lot more beyond that so I cannot guarantee that some other action does not create other files that need to be deleted.

precisecalc.txt
HBelusca: Rapps uses the standard windows ini parser, which understands the external quotes.
See https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilestring#remarks
the third paragraph there.

restoration.txt
julenuri: It seems to be a weird old clumsy software/abandonware, and majorgeeks it's not an official source, neither the softpedia, neither the one I upload here. It's hard to find the original source. I will try to find that one.
https://web.archive.org/web/20210328174143/www3.telus.net/mikebike/Restoration.zip

multimc.txt
```diff
@@ -6,7 +6,14 @@ Description = MultiMC is a custom launcher for Minecraft that allows you to have
 Category = 4
 URLSite = https://multimc.org/
 URLDownload = https://files.multimc.org/downloads/mmc-develop-win32.zip
-SHA1 = 94b8e884cbf59378754bc300e60e570ed204e819
-SizeBytes = 14141416
+SHA1 = 1096e20d0c8f1abe61d00596ed166843833c84d9
+SizeBytes = 14255069
```
julenuri, line 10: If the hash changed so much, isn't the version changing too?

Folder2Iso.txt
julenuri: Actually, the original .zip have both binaries, x86 and x64, and the adding the Amd64 entry make it to appear. If we want to make it appear as a setup, we need to discriminate the setup by the architecture, or better this that is an easier solution.

julenuri (changes requested):
Minor tiny edits, I am okkay for all the other files.

xbgpack1.txt
whindsaks: 1. It points to archive.org, not the chat.
2. I did not change this url in this PR.

Folder2Iso.txt
whindsaks: > Actually, the original .zip have both binaries, x86 and x64

Which "original .zip" are you talking about now? The one we are currently using (for x86) only contains the AMD64 binary. The author does not provide a usable direct link we can use to fix this.

multimc.txt
whindsaks: The exe does not have proper version information but the about dialog indicates the build number has changed, yes. I will update.

Folder2Iso.txt
julenuri: Okkay, don't worry, I will fix that situation in a newer PR.

julenuri (approved):
In terms of PR adequacy, approved. 
In terms of testing, I have still to test the behavior of all the packages.

HBelusca (approved):
LGTM on the technical aspect. Thanks for your exhaustive work!

PR #293, opened 2025-01-22
[RAPPS-DB] Add generic wallpaper icon for wallpaper packs

whindsaks:
The icon is a modified version of Tango preferences-desktop-wallpaper.


roswallp.txt
```diff
@@ -6,12 +6,14 @@ Description = The standard wallpaper collection for ReactOS
...
 URLDownload = https://github.com/reactos/wallpapers/releases/download/v3/wallpapers.zip
-Installer = Generate
 SizeBytes = 7474167
+Icon = wallpaper.ico
+Installer = Generate
 
 [Generate]
 Files = *
 Dir = %SystemRoot%\Web\Wallpaper
+Icon = ..\..\system32\desk.cpl
```
HBelusca, line 16: Why the need for this?

HBelusca:
What are the sizes of the wallpaper.ico ?

whindsaks:
> What are the sizes of the wallpaper.ico ?

32x32 and 48x48

HBelusca:
> > What are the sizes of the wallpaper.ico ?
> 
> 32x32 and 48x48

Maybe add 16x16 too?

julenuri (approved):

roswallp.txt
HBelusca: What about `%SystemRoot%\system32\desk.cpl` then? It would look a bit less hacky.

whindsaks:
> Maybe add 16x16 too?

Rapps uses the normal Windows icon size for the icons in the app list, this will never be 16x16? I'll add it anyway.

roswallp.txt
whindsaks: No, that icon is relative to the install directory.

Rapps could perhaps support and use special handling for paths starting with % for the icon but it currently does not.

HBelusca: I was suggesting something like what's used in the Dir syntax (and basically, if whatever is given there as the value was passed to some PathCanonicalize function, maybe it would understand that after Expanding the environment strings in "install_dir\%SystemRoot%\whatever", the thing would actually refer to the systemroot\whatever.
(I hope you understand what I mean here ^^)

HBelusca (approved):

roswallp.txt
whindsaks: Without looking it up, I believe Rapps calculates the icon path as something like `CStringW icopath = PathAppend(installdir, GetIniStr("Icon"));` (pseudo code).

We could check if the first character is % and then don't prefix the install directory but the current code in master cannot handle this AFAIK.

oleg-dubinskiy:
> > Maybe add 16x16 too?
> 
> Rapps uses the normal Windows icon size for the icons in the app list, this will never be 16x16? I'll add it anyway.

They were 16x16 until 2020, when I've sent a #3144 PR with implementation of displaying icons for installed apps in Rapps. It was required because usually apps contain 32x32 icons those are defined in their registry entries ("Uninstall" section), and they were displayed ugly with 16x16 size. See https://github.com/reactos/reactos/pull/3144#discussion_r486273471.

oleg-dubinskiy (approved):

PR #297, opened 2025-01-30
[RAPPS-DB] Convert more zip packages to zip installers

whindsaks:
Notes:
- ccleste, cstalin, fall, maelstorm and xrick rely on the recently added Rapps shortcut working directory enhancement to run correctly.
- CStalin defaults to Spanish but that is a flaw in their .zip distribution and not something this PR can fix.
- Neverball should probably be converted at some point but I could not get it to run in my VM.
- Openarena should probably be converted at some point but it's so large, I did not test it.

### DIFF
--- a/hiawatha.txt
+++ b/hiawatha.txt
@@ -8,6 +8,14 @@ URLSite = https://www.hiawatha-webserver.org/
 URLDownload = https://web.archive.org/web/2024if_/https://www.hiawatha-webserver.org/files/hiawatha-9/hiawatha-9.15.zip
 SHA1 = 6ada69afdeb3f75ceda0c19c67d562a5052a9949
 SizeBytes = 3560500
+Installer = Generate
+
+[Generate]
+Files = Hiawatha\Hiawatha.bat|*.txt|Hiawatha\*
+DelFile = work\hiawatha.pid
+DelDir = logfiles
+DelDirEmpty = work\upload|work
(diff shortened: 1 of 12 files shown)


hiawatha.txt
```diff
@@ -8,6 +8,14 @@ URLSite = https://www.hiawatha-webserver.org/
...
 SHA1 = 6ada69afdeb3f75ceda0c19c67d562a5052a9949
 SizeBytes = 3560500
+Installer = Generate
+
+[Generate]
+Files = Hiawatha\Hiawatha.bat|*.txt|Hiawatha\*
+DelFile = work\hiawatha.pid
+DelDir = logfiles
+DelDirEmpty = work\upload|work
+DelRegEmpty = HKCU\Software\Cygwin\Program Options|HKCU\Software\Cygwin
```
HBelusca, line 18: Does this work fine in case there exists other Cygwin-based programs (incl. Cygwin itself) on the system?

HBelusca (approved):

hiawatha.txt
whindsaks: Yes because it only deletes if the key is empty (hiawatha does not seem to write anything there)

julenuri (approved):
Pending of testing this under ReactOS to see if the behavior is the expected.

whindsaks:
> Pending of testing this under ReactOS to see if the behavior is the expected.

I personally tested every installer changed here but of course mistakes and oversights can happen.

PR #298, opened 2025-02-10
[RAPPS-DB] Update RosBE, WinSpy and minor fixes

whindsaks:
RosBE has to be updated because the old version does not work on current master "(CMake Error at CMakeLists.txt:2 (cmake_minimum_required): CMake 3.17.0 or higher is required.)".

I have verified that the new version successfully installs on 0.4.14, can `configure.cmd&cd output-MinGW-i386&ninja winver` successfully on current master in 0.4.14 (and in 0.4.16 of course) and can `configure.cmd&cd output-MinGW-i386&ninja winver` successfully on reactos-0.4.14-release.zip.

----

I'm adding a newer WinSpy++ version for x64 but keeping the old one for x86 because the new version uses layered windows and ROS does not fully support those. There is no older x64 version we can use.

### DIFF
--- a/deathrally.txt
+++ b/deathrally.txt
@@ -11,4 +11,4 @@ URLDownload = https://web.archive.org/web/201502im_/http://gamedaily.newaol.com/
 SHA1 = bda2a2230c01580ac779f3cf0f5954d234c8c78e
 SizeBytes = 43574432
 SaveAs = DeathRallyWin_10.exe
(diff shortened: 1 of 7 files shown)


learn-more (approved):

julenuri (approved):
Good code cleaning and adding versions. I want to test that RosBE 2.2.1 in current ReactOS code, I remember I did in the past, but... I wonder if it's still possible today. :+1:

PR #303, opened 2025-04-04
[RAPPS-DB] Add TweakUI and extra MS visual styles

whindsaks:
Notes:
- Royale/Energy Blue is the one released by Microsoft New Zealand that includes the new bliss wallpaper (New Bliss.jpg). The other version _("Microsoft Windows XP Tablet PC Edition 2005 Energy Blue Theme Pack")_ (with EnergyBliss.jpg) is available [here](https://web.archive.org/web/20051124174056/http://download.microsoft.com/download/5/2/9/5295fa0c-575a-4ee6-b186-0a8cf7ddfde6/WindowsXP-TabletPC-EnergyBlueTheme-x86-ENU.exe) but will only install on XP Tablet PC edition.
- [Royale Noir](https://www.betaarchive.com/wiki/index.php/Windows/Visual_Styles/Royale_Noir) is signed by Microsoft and works on stock Windows XP but I'm not sure where it was initially released, probably somewhere on Channel 9. The download points to a .zip with the name fixed from Luna to Royale Noir so it does not conflict with the real Luna.
- The Zune theme installer requires the user to set compatibility settings on the .msi. Perhaps Rapps could gain the ability to do this automatically in the future?
- Neither version of TweakUI works perfectly so I included both.
- TweakUI v2 will not uninstall because it uses a custom MSHTA uninstaller that needs Trident+WScript.

### DIFF
--- a/royaletheme.txt
+++ b/royaletheme.txt
@@ -1,0 +1,11 @@
+[Section]
+Name = Royale
+Publisher = Microsoft
+License = Freeware
+LicenseType = 2
+Description = Royale (also known as Energy Blue and Media Center style) is a visual style designed for Windows XP Media Center Edition 2005. Also includes the New Bliss wallpaper.
+Category = 15
+URLSite = https://web.archive.org/web/20051028063258if_/http://www.microsoft.com/nz/windowsxp/downloads/nzbliss.aspx
+URLDownload = https://web.archive.org/web/20051128025233im_/http://download.microsoft.com/download/a/c/4/ac407ba2-30d2-447b-8c93-83967a92dd9f/royale.zip
+SHA1 = 08d5face1bbe9cde6944b8b5ae453fe2a117faec
+SizeBytes = 688390
--- a/royalenoirtheme.txt
+++ b/royalenoirtheme.txt
@@ -1,0 +1,16 @@
+[Section]
+Name = Royale Noir
+License = Freeware
+LicenseType = 2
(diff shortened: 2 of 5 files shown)


Doug-Lyons (approved):

julenuri (approved):

royaletheme.txt
```diff
@@ -0,0 +1,11 @@
+[Section]
+Name = Royale
+Publisher = Microsoft
+License = Freeware
+LicenseType = 2
+Description = Royale (also known as Energy Blue and Media Center style) is a visual style designed for Windows XP Media Center Edition 2005. Also includes the New Bliss wallpaper.
+Category = 15
+URLSite = https://web.archive.org/web/20051028063258if_/http://www.microsoft.com/nz/windowsxp/downloads/nzbliss.aspx
+URLDownload = https://web.archive.org/web/20051128025233im_/http://download.microsoft.com/download/a/c/4/ac407ba2-30d2-447b-8c93-83967a92dd9f/royale.zip
```
HBelusca, line 9: That one didn't require the special installer thing like for Royale Noir?

HBelusca (approved):

royaletheme.txt
whindsaks: This one is a nasty exe installer inside a zip that we don't have a good experience for (fix for this [here](https://github.com/reactos/reactos/pull/7866)). I left it like this since this is the original file on a Microsoft server.

Royale Noir just has the style files directly in the .zip so we can use our Generate installer feature...

PR #305, opened 2025-04-29
Use ExeInZip automation on relevant packages

whindsaks:
Use ExeInZip in the cases where we can automatically run the setup application inside a zip package. Also fixes the Fotografix download and some other minor attributes.


julenuri (changes requested):
To add the Fotografix fix in the commit message and for me is a ship it right now option.

julenuri (approved):

PR #309, opened 2025-07-05
[RAPPS-DB] Fix theming/wallpapers to be more consistant in the database.

julenuri:


julenuri:
> - Why did you move wallpapers into the themes category?
Because they don't fit in other categories the way they fit into Theming. They are not tools and neither Others. Because it's a aesthetic/theming/wallpaper feature.
> - Why did you give both visual styles and wallpapers the same icon?
I don't have any other Tango functional icon in hand. If there are alternative icons that are interesting, just ping me or commit it.
> It's really annoying to tell wallpapers and visual styles apart now!
Well, It's way easier to find them than in the Others category. If we want different icons, ask me, I will find some.

Happens the same with the IME and KJC fonts. They don't fit exactly in categories.

whindsaks:
> I don't have any other Tango functional icon in hand.

I would personally just use oldinstaller.ico. If you want something better, the desk.cpl icon is slightly different from the wallpaper icon. (Install the Royale Noir theme and look in the uninstall section of Rapps to see it).

PR #310, opened 2025-07-05
[RAPPS-DB] dding icons to the MS files, plus others and fix MSVC 2017 because they updated the file (again)

julenuri:
Adding icons for some MS files.


julenuri:
@whindsaks yep, I can make improvements to these icons. Actually I took nearly the original ones.

PR #320, opened 2025-07-10
[RAPPS-DB] Again, minor fixes and links that I know that are broken or outdated.

julenuri:


krita.txt
```diff
@@ -14,6 +14,12 @@ Version = 3.2.1
 URLDownload = https://download.kde.org/Attic/krita/3.2.1/krita-3.2.1-x64.zip
 SHA1 = cb85d630a8efdbc816296d757e05b1dc3d77943d
 SizeBytes = 100532224
+Installer = Generate
+
+[Generate]
+Files = *.exe|*.*
+Dir = %ProgramFiles%\Krita
+Lnk = "Krita.lnk"
```
whindsaks, line 22:
~~~suggestion
~~~

Other than programfiles which I don't understand why you are forcing, the other values are the defaults and you don't need to add them.

chrome_48.txt
```diff
@@ -14,6 +14,12 @@ Icon = chrome.ico
 URLDownload = https://archive.org/download/48.0.2564.109-chrome-installer/chrome48_x64.zip
 SHA1 = 18a47a2285f05665e3ad19446186cfbcc30e5ba3
 SizeBytes = 80979795
+Installer = Generate
```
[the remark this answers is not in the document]
julenuri: Exactly the same parameter that needs the x86 version of Chromium and Chrome.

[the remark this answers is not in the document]
whindsaks: I don't know what that parameter is. But I can add the feature to Rapps.

krita.txt
julenuri: Krita might need the program files directory. Also, users will be confused if it's software is not in the directory that is expected. But if you consider contrary, I can commit it your suggestion.

chrome_48.txt
```diff
@@ -14,6 +14,12 @@ Icon = chrome.ico
...
 
 [Section.0a]
 Description = Google Chrome es el popular navegador web de código cerrado​ desarrollado por Google, derivado del proyecto Chromium.
+
+[Generate]
+Files = *.exe|*.*
+Dir = %ProgramFiles%\Google\Chrome
```
whindsaks, line 24: Use the defaults?

chrome_48.txt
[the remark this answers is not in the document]
julenuri: I mean, --no-sandbox it's compulsory in every Chromium based engine.

krita.txt
whindsaks: If the app works if you don't force the directory, please don't force. If you need to force the directory, fine. But then you should also set Scope = machine in [Section] (This does not actually do anything but will if we ever get UAC support).

julenuri: Okkay.

coretemp.txt
```diff
@@ -13,7 +13,10 @@ SizeBytes = 1294336
 Version = 1.18.1 Portable
 URLDownload = https://www.alcpu.com/CoreTemp/CoreTemp64.zip
 SHA1 = f588216efeebfd7e4012ce5346f15cd4c0c81dbc
-SizeBytes = 458752 
+SizeBytes = 458752
```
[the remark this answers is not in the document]
julenuri: It had one space after the number before. I deleted it. Pointless, but...

PR #325, opened 2025-07-19
Fixed multiMC download.

Coolgamerguy-tech:
MultiMC download link was to a URL that changed the file for new versions, which made checking the file against the SHA-1 fail.
I also deleted minecraft 4K as the link was broken (and not on archive.org) and it is included in MultiMC already.
Screenshot 2025-07-19 125106


julenuri (approved):
Okkay from my side.

PR #327, opened 2025-08-06
[RAPPS-DB] Applications for ARM and some updates of relevant packages (Part I)

julenuri:
I will commit in this PR multiple versions of packages that I consider relevant or important for ReactOS.

Consider that this PR also adds a huge number of ARM builds, that should be tested. Logically, not in ReactOS (yet ™) but obviously in Windows on ARM devices, Windows RT, or whatever uses ARM and boots NT.

Some of the repos of those files are official. But sadly, some of them are built by developers or users of WoA devices, so... To be consider a decent test and revision before commit anything.

Some-when, in the future, this PR will be relevant for the ReactOS AArch32 and Aarch64 ports. Or maybe, only supports Surface users. I would consider both of them like a pass.

I was thinking on splitting it, but once I started, well this went huge, and fixed some .txt in the way, like opera.

If you want better options for binaries, comment and I will compare them. Thanks.

PD: Lord, this is a nightmare.
PD2: Still WIP.

### DIFF
--- a/audacity.txt
+++ b/audacity.txt
@@ -26,12 +26,22 @@ SHA1 = 6cded0e7f2c50c87aa6fb5e0385ffef6f16d093e
 SizeBytes = 4725083
 Installer = Generate
 
+[Generate]
+Files = Audacity\Audacity.exe|Audacity\*.dll|Audacity\license\*.txt
+Icon = Audacity.exe,4
+DelReg = HKCU\Software\ExampleSoft\SillyGame
+DelRegEmpty = HKCU\Software\ExampleSoft
+Lnk = Audacity.lnk
+
 [Section.arm64]
 Version = 2.4.2
 URLDownload = https://github.com/henricj/audacity/releases/download/latest-arm64/MSI-Audacity-arm64-Audacity-2.4.2-84-gfc46c01f3.zip
 SHA1 = 4a28e28ae58a9b5d562fc44abcec6b42fd4927b9
 SizeBytes = 23110953
 Installer = ExeInZip
+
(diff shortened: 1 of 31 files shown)


putty.txt
```diff
@@ -1,19 +1,28 @@
...
-Version = 0.80
+Version = 0.83
 License = MIT
 Description = A free SSH, Telnet, rlogin, and raw TCP client.
 Category = 5
 URLSite = http://www.chiark.greenend.org.uk/~sgtatham/putty/
-URLDownload = https://the.earth.li/~sgtatham/putty/0.80/w32/putty-0.80-installer.msi
-SHA1 = 84b295c34fb91f073331795f9db69c6465ded064
-SizeBytes = 3419136
+URLDownload = https://the.earth.li/~sgtatham/putty/0.83/w32/putty-0.83-installer.msii
```
whindsaks, line 8: .msii typo?

julenuri:
imagen
It's really practical, but I have to improve some entries. It's not perfect, yet.

audacity.txt
```diff
@@ -29,8 +29,6 @@ Installer = Generate
 [Generate]
 Files = Audacity\Audacity.exe|Audacity\*.dll|Audacity\license\*.txt
 Icon = Audacity.exe,4
-DelReg = HKCU\Software\ExampleSoft\SillyGame
-DelRegEmpty = HKCU\Software\ExampleSoft
```
whindsaks: You can also remove Lnk =

whindsaks (approved):

PR #331, opened 2025-08-16
Change the luna-beta's link to the original URL of the repo (ít works)

itsmengwh08:
Since the original link to the repo still works fine, I'll leave it here.


itsmengwh08:
I think the original link going straight to the original repo is better than struggling with the Wayback Machine

PR #384, opened 2026-04-12
[RAPPS-DB] Add software icons from A-Z and little fix about windir

julenuri:
Make the list a little bit more complete, little by little. 

:D


learn-more (changes requested):
These icons are a fair bit bigger than others, maybe we can cut out the png images (assuming they have those).


doublecommander.ico
everedit.ico‎
freestopwatch.ico
npackd.ico‎
pdfsam.ico‎
ptanks.ico
remood.ico‎



Maybe it would be something for the build validation to check for that?

julenuri:
@learn-more It's a good idea. Also... We should set a limit of the icon size maybe? Like n, which can be a number between 0 and 30KB, not 100% sure about this statement tho, and having some strip down version of some .ico. All of the quoted have multiple low res version, high res, big size. I can't tell...



But first, I want to complete the list fully. Later we will talk about the size.

julenuri:
imagen
Those are the biggest .ico files

julenuri:
Once I strip down the files to the 100KB number, I will commit it if you agree.

whindsaks:
> Once I strip down the files to the 100KB number

Compare this to what I did with the recently added alex4.ico, I used an optimized palette with 16 colors (IcoFX in RAPPS has a menu item to convert to an optimized palette of x colors). Clearly not all icons can accept such treatment but most can at least be reduced to 256 colors (except the 16x16 sub images, these are actually smaller when you make them 24/32-bit color rather than 256).

julenuri:
> > Once I strip down the files to the 100KB number
> 
> Compare this to what I did with the recently added alex4.ico, I used an optimized palette with 16 colors (IcoFX in RAPPS has a menu item to convert to an optimized palette of x colors). Clearly not all icons can accept such treatment but most can at least be reduced to 256 colors (except the 16x16 sub images, these are actually smaller when you make them 24/32-bit color rather than 256).

Yes, I looked at it. I am not concern about how they look right now in ReactOS yet, I haven't looked at it.

I tried with ArtIcons, and was really fast to do. And the filesize suffers more for the dimensions more than the color bit, and having them duped without using them like none of the times, has no sense. 

Some of the icons have been stripped down from 200KB to 32KB or less. More reasonable. Some of them can be more stripped down, but I will maintain it like this for this PR. This is only for upgrading the existing icons and adding the lost ones.

whindsaks:
> ArtIcons

ArtIcons can't do 16 color optimized palette (it can't create/convert them but it can read them, it only converts 16 colors forced to the standard colors) but is otherwise pretty good.

julenuri:
> > ArtIcons
> 
> ArtIcons can't do 16 color optimized palette (it can't create/convert them but it can read them, it only converts 16 colors forced to the standard colors) but is otherwise pretty good.

I will have to give IcoFX a look in the next PR ;)

whindsaks:
> I will have to give IcoFX a look in the next PR ;)

ArtIcons has much better handling of PNG compression, you can turn it on/off on specific sub-images, IcoFX just automatically compresses the 256x256 image (IIRC). I use both editors.

whindsaks (approved):

Doug-Lyons (approved):

PR #386, opened 2026-05-14
[RAPPS-DB] Dependencies

whindsaks:
Use the new dependencies feature from [ROS PR 8963](https://github.com/reactos/reactos/pull/8963) for packages that need it.

### DIFF
--- a/dvdwritenow.txt
+++ b/dvdwritenow.txt
@@ -2,7 +2,7 @@
 Name = DVD Write Now
 Version = 1.5.12 SP2
 License = Freeware
-Description = A free direct CD/DVD recording software. Once installed, you can use your CD/DVD RW as USB Flash or removable hard drives. Along with this you can write to CD/DVD RW recorded with UDF filesystem. You can also read UDF disks created with other software (Microsoft DVD−ROMs, RW disks recorded with Ahead InCD). You need a ZIP decompression program to install it.
+Description = A free direct CD/DVD recording software. Once installed, you can use your CD/DVD RW as USB Flash or removable hard drives. Along with this you can write to CD/DVD RW recorded with UDF filesystem. You can also read UDF disks created with other software (Microsoft DVD−ROMs, RW disks recorded with Ahead InCD).
 Category = 12
 URLSite = https://dwn.alter.org.ua/
 URLDownload = https://web.archive.org/web/20150606193303if_/http://dwn.alter.org.ua/downloads/dwn_1_5_12_sp2.zip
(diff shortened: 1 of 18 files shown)


vc2008sp1run.txt
```diff
@@ -9,6 +9,14 @@ URLDownload = https://web.archive.org/web/20200819111521/http://download.microso
 SHA1 = 470640aa4bb7db8e69196b5edb0010933569e98d
 SizeBytes = 4479832
 Icon = oldwinsetup.ico
+SilentParameters = /q /norestart
+
+[Section.x86]
+RegName = {9BE518E6-ECC6-35A9-88E4-87755C07200F}
+
+[Section.amd64]
+URLDownload = https://download.microsoft.com/download/5/D/8/5D8C65CB-C849-4025-8E95-C3966CAFD8AE/vcredist_x64.exe
```
katahiromz, line 18: Is SHA1 for amd64 needed?

whindsaks: The hash is technically optional. I was lazy and trust https from Microsoft. I got these from a list on a webpage, I did not download them.

binarymaster: I'm not sure RAPPS would resist a MITM attack without a hash. 🤔
(didn't tested that scenario though)

katahiromz: And how about `SizeBytes`?

whindsaks: Sizebytes is for display in the UI. This PR is really for fixing the packages that have dependencies, not the dependencies themselves. There are more VC runtime packages that could use some updates but I'm not touching them at all because no packages depend on them...

katahiromz (approved):

vc2008sp1run.txt
katahiromz: Wrong display OK?

katahiromz (approved):
Ok.

dvdwritenow.txt
```diff
@@ -2,7 +2,7 @@
 Name = DVD Write Now
 Version = 1.5.12 SP2
 License = Freeware
-Description = A free direct CD/DVD recording software. Once installed, you can use your CD/DVD RW as USB Flash or removable hard drives. Along with this you can write to CD/DVD RW recorded with UDF filesystem. You can also read UDF disks created with other software (Microsoft DVD−ROMs, RW disks recorded with Ahead InCD). You need a ZIP decompression program to install it.
+Description = A free direct CD/DVD recording software. Once installed, you can use your CD/DVD RW as USB Flash or removable hard drives. Along with this you can write to CD/DVD RW recorded with UDF filesystem. You can also read UDF disks created with other software (Microsoft DVD−ROMs, RW disks recorded with Ahead InCD).
 Category = 12
 URLSite = https://dwn.alter.org.ua/
 URLDownload = https://web.archive.org/web/20150606193303if_/http://dwn.alter.org.ua/downloads/dwn_1_5_12_sp2.zip
```
julenuri, line 8: Is it setuping itself after decompress? Or needs a generate command?

vc6run.txt
```diff
@@ -2,13 +2,14 @@
...
 License = EULA
-Description = Needed by applications compiled with Visual Studio 6. Contains: advpack.dll, asycfilt.dll, atla.dll, atlu.dll, comcat.dll, mfc42.dll, mfc42u.dll, msvcirt.dll, msvcp60.dll, msvcrt.dll, oleaut32.dll, olepro32.dll.
+Description = Needed by applications compiled with Visual Studio 6. Contains: advpack.dll, asycfilt.dll, atla.dll, atlu.dll, comcat.dll, mfc42.dll, mfc42u.dll, msvcirt.dll, msvcp60.dll, msvcrt.dll, oleaut32.dll, olepro32.dll. The inner vcredist.exe must have its compatibility set to NT4SP5.
 Category = 14
 URLSite = https://support.microsoft.com/kb/259403/
 URLDownload = https://download.microsoft.com/download/vc60pro/Update/2/W9XNT4/EN-US/VC6RedistSetup_deu.exe
 SHA1 = a8c4dd33e281c166488846a10edf97ff0ce37044
 SizeBytes = 1837888
 Icon = oldwinsetup.ico
+SilentParameters = /Q /R:N
```
julenuri, line 12: So now it installs silently without failing because the reported version? Or will it fail until we set the NT4SP5?

julenuri (approved):
I am happy with this PR. 
Not 100% sure if all the dependencies are covered but it's a nice start.

dvdwritenow.txt
whindsaks: It already uses the ExeInZip feature, we just forgot to update the description.

vc6run.txt
whindsaks: The silent switch only applies to the outer extractor (skips the license page), you still need to set NT4SP5 on the inner extractor manually. (I made a NSIS installer wrapper that can can do all this automatically and silently, we just need someone to upload it to svn.reactos.org/packages)

whindsaks:
> Not 100% sure if all the dependencies are covered but it's a nice start.

I also don't know if I got them all, I mainly just looked at the package descriptions.

vc6run.txt
julenuri: Good to know.

glidewrapzbag.txt
```diff
@@ -8,6 +8,7 @@ URLSite = https://www.zeckensack.de/glide/
 URLDownload = https://web.archive.org/web/20250414094032/zeckensack.de/glide/archive/glidewrapper084c.exe
 SHA1 = 7a9d60a18b660473742b476465e9aea7bd5ab6f8
 SizeBytes = 254995
+RegName = GlidewrapZbag
```
HBelusca, line 11: @whindsaks could you explain me what the purpose of RegName is (since it's not systematically present for all packages)?

HBelusca (approved):

glidewrapzbag.txt
whindsaks: By default we assume that the apps registry uninstall subkey has the same name as the package display name. In the cases where it does not, [RegName should be set in the package manifest](https://reactos.org/wiki/RAPPS#RegName). Rapps uses this to detect if the application is installed or not (`CAvailableApplicationInfo::IsInstalled` and `CAvailableApplicationInfo::InsertVersionInfo`).

To find an apps "RegName", just install the app and look in the registry...

PR #388, opened 2026-05-24
fix or add some sha1 hashes

xlucn:
There are
- 1 mismatched SHA1 fixed. No idea why it was different before.
- 4 missing SHA1 added. It is optional, but why not add it :)


katahiromz (approved):

whindsaks (approved):

julenuri (approved):
I am guessing Fopnu and Tixati are going to be continuously suffering the diff in the hash. 😢

learn-more:
> I am guessing Fopnu and Tixati are going to be continuously suffering the diff in the hash. 😢

Then we should pin it via web.archive.org

PR #390, opened 2026-06-08
[RAPPS-DB] Add UT and Alex4 games (minor mesa and vc2015 edit)

whindsaks:
Apparently Unreal and UT are [free now.](https://www.pcgamer.com/games/fps/unreal-gold-and-unreal-tournament-are-now-free-on-the-internet-archive-and-epic-says-thats-a-okay/)

Notes:
 - I faked the UT size, it will download about 700mb when you run the installer.


katahiromz (approved):

unrealtournament.txt
```diff
@@ -0,0 +1,12 @@
+[Section]
+Name = Unreal Tournament: Game of the Year Edition
+LicenseType = 2
+Description = (Choose "Direct3D 9" and "Cursor input") An arena first-person shooter, with head-to-head multiplayer deathmatches.
+Category = 4
+Publisher = Epic Games
+Screenshot1 = https://web.archive.org/web/20120602003535im_/http://upload.wikimedia.org/wikipedia/en/f/f4/Unrealtournament.jpg
+URLDownload = https://github.com/OldUnreal/FullGameInstallers/releases/download/windows-game-installers/UT_GOTY.exe
+SHA1 = d22350da85cbccebd142a93d86873417b9c11c3e
+SizeBytes = 787654321
```
HBelusca, line 10: Could you explain a bit more why it was needed to fake the dl size?

HBelusca (approved):

unrealtournament.txt
whindsaks: The initial download of UT_GOTY.exe is 90mb but it is going to download about 700mb more as its first step when you run it. I don't **need** to fake it, I'm just doing it to give the user a better sense of how big the download actually is.

HBelusca: Ok; but would this affect the download gauge (in the download dialog), or is this value purely informative (and is used only in the  description text zone)?

whindsaks: It's also used in the download dialog as a hint but as soon as it gets the response from the server, it gets reset to the "real" size.

julenuri (approved):
An obviously yes for my side.

PR #391, opened 2026-07-03
[RAPPS-DB] Update 7-Zip

whindsaks:


windirstat.txt
```diff
@@ -1,7 +1,8 @@
 [Section]
 Name = WinDirStat
 Version = 1.1.2
-License = GPL
+License = GPLv2
+LicenseType = 1
```
HBelusca, line 5: Maybe make a separate PR just for that one (it's not 7zip).

HBelusca (approved):
One remark, but OK otherwise.

windirstat.txt
whindsaks: See that the branch name is Appupdates26Jul. I had some other apps to update but it turns out that the new versions don't actually work on ROS so I removed them again from my branch.

julenuri (approved):

PR #394, opened 2026-07-14
Update mypal

ghost:


oleg-dubinskiy:
@Coolgamerguy-tech did you test it in ReactOS? Does it work stable enough? Do you have a screenshot of it working in ReactOS? As I remember, it still fails when opening the browser menu from top right (because of layered windows code, ROS bug). Also, UI may hang sometimes when loading web-pages, so it's too early to update it I think.

ghost:
> @Coolgamerguy-tech did you test it in ReactOS? Does it work stable enough? Do you have a screenshot of it working in ReactOS? As I remember, it still fails when opening the browser menu from top right (because of layered windows code, ROS bug). Also, UI may hang sometimes when loading web-pages, so it's too early to update it I think.

oops, you are correct, opening the menu crash, sorry.

PR #399, opened 2026-07-28
[RAPPS-DB] Use `SaveAs` to avoid VC and .NET installer name collisions

xlucn:
**Problem:**
I noticed some downloaded files have the same filename. For example, `vc2008sp1run` and `vc2010sp1run` are both named `vcredist_x86.exe`. That results in the files of different versions replacing each other after download, so those files can't be kept, even if the user wants to keep them for possible future reinstall.

**Solution:**
The [`SaveAs`](https://reactos.org/wiki/RAPPS#SaveAs) fields are added so that the downloaded files are properly renamed to avoid filename collisions:
- For Microsoft Visual C++ Redistributables, a `_vs20xx` is added before the extension, following the naming convention found in `vcredist_arm_vs2015.exe`
- For .NET Framework v1.1 and v2.0, `11` or `20` is added before the extension, following the naming convention of the app name.

Not sure if this is the intended usage of `SaveAs`. It solves the problem tho :)

**Test:**
I have tested locally in applications manager by updating the local appdb. The files are saved in different filenames, and can be installed with, as expected.

### DIFF
--- a/net11.txt
+++ b/net11.txt
@@ -8,6 +8,7 @@ URLSite = https://www.microsoft.com/en-US/download/details.aspx?id=26
 URLDownload = https://web.archive.org/web/20210505032023if_/http://download.microsoft.com/download/a/a/c/aac39226-8825-44ce-90e3-bf8203e74006/dotnetfx.exe
 SHA1 = 16a354a2207c4c8846b617cbc78f7b7c1856340e
 SizeBytes = 24265736
+SaveAs = dotnetfx11.exe
 Icon = oldwinsetup.ico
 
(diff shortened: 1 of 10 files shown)


binarymaster (approved):

whindsaks:
> Not sure if this is the intended usage of `SaveAs`.

It was designed for URLs where it does not have .exe or anything sensible in the name but your usage is a nice tweak.

whindsaks (approved):

julenuri:
Good point, really. 👍

julenuri (approved):

PR #403, opened 2026-08-29
[RAPPS-DB] Add Xash3D FWGS engine

PavelCheloveckov:
Xash3D FWGS - Compatibility engine for Half-Life and old GoldSrc games
Tested on ReactOS 0.4.16 environment.

image


xash3d.txt
```diff
@@ -0,0 +1,21 @@
+[Section]
+Name = Xash3D FWGS
+Version = continuous-master
+License = GPLv3
+Description = Open-source compatibility engine for Half-Life and old GoldSrc games.
+Category = 4
+URLSite = https://github.com/FWGS/xash3d-fwgs
+URLDownload = https://github.com/FWGS/xash3d-fwgs/releases/download/continuous/xash3d-fwgs-win32-i386.7z
```
learn-more, line 8: this uri sounds like the contents is going to change, is it possible to pin a version here?

PavelCheloveckov: > this uri sounds like the contents is going to change, is it possible to pin a version here?

There is a version 0.20, but it is no longer relevant (since it was released in 2019). So I can just remove sha1 and SizeBytes.

julenuri: > > this uri sounds like the contents is going to change, is it possible to pin a version here?
> 
> There is a version 0.20, but it is no longer relevant (since it was released in 2019). So I can just remove sha1 and SizeBytes.

We want the SHA1 + SizeBytes. That could provide us a little security about what we download. Also, I am not sure if the rapps-db admits to don't have the entry.

julenuri:
I have a deep internal debate to approve or not this PR. 
In one side, it's an interesting project of an open sourced famous videogame engine. Moreover it is in active development and it's compatible with ReactOS. That's great.
But in the other side, I have doubts about to add it. You still need the assets and some files of the original and breaks the "out of the box experience" of the rapps-db base. You will click to download and you will lack files to make it run, and we won't provide them...
Correct me if I am wrong, because I haven't investigated or tested the game enough, but if the statement is like that... I highly doubt to approve it.

PavelCheloveckov:
> I have a deep internal debate to approve or not this PR. In one side, it's an interesting project of an open sourced famous videogame engine. Moreover it is in active development and it's compatible with ReactOS. That's great. But in the other side, I have doubts about to add it. You still need the assets and some files of the original and breaks the "out of the box experience" of the rapps-db base. You will click to download and you will lack files to make it run, and we won't provide them... Correct me if I am wrong, because I haven't investigated or tested the game enough, but if the statement is like that... I highly doubt to approve it.

@julenuri Yes, it requires original game files to run, but that's a standard practice for all source ports (like Doom, Quake or OpenMW) in package managers. Users just copy their own game data into the installation folder. It's a great addition for the ReactOS gaming ecosystem. If needed, I can simply add a note to the description field that original game assets are required. What do you think?

PR #407, opened 2026-09-04
[RAPPS-DB] Move Notepad++ from Office to Development category

jeditobe:
Notepad++ is obviously not an office app! Six-seven!


binarymaster (approved):

whindsaks:
Or is the generic tool category a better fit? Sysadmins etc. will edit configuration files.

HBelusca (approved):
It technically depends for what purposes someone used npp.

HBelusca:
Side thought: I wonder whether we support showing a given app in more than one single category.

whindsaks:
> I wonder whether we support showing a given app in more than one single category

We don't. If we are lucky, the current version will be able to parse 1|7 as 1 without errors, then there is a chance.

jeditobe:
Notepad++ should be in the same category as Notepad2e, that is the idea

learn-more (approved):

whindsaks:
> If we are lucky, the current version will be able to parse 1|7 as 1 without errors

And we are lucky, current RAPPS uses `CConfigParser::GetInt` and that just calls `StrToIntW` which just stops parsing after the first non-numeric character without errors. So we can extend this in the future to support multiple categories per package...

whindsaks:
> in the future to support multiple categories per package

[RappsAppMultiCat](https://github.com/reactos/reactos/compare/master...whindsaks:reactos:RappsAppMultiCat)

